PR #1, opened 2016-04-20
Rest api

sztamas:
@id2359 @mradochonski can you please have a look/review my changes?
It might be also useful for you guys to check out the rest_api branch and then just start exploring from /api/v1/.

You will see the endpoints in there + I made some extra effort to make the API more browsable. For example you go to registries, you will have patients_url and clinitian_url which are endpoint for patients/clinitians in a given registry.

I'm also open to meet over video conference. Maybe that will be the easiest for me to take you through the changes I've made, you ask questions, make suggestions etc.

Thanks,

Tamas


rdrf/rdrf/api_urls.py
```diff
@@ -0,0 +1,23 @@
...
+
+router = DefaultRouterWithSimpleViews()
+router.register(r'registries', api_views.RegistryList, base_name='registry')
+router.register(r'users', api_views.CustomUserViewSet)
+router.register(r'workinggroups', api_views.WorkingGroupViewSet)
+router.register(r'countries', api_views.ListCountries, base_name='country')
+router.register(r'genes', api_views.LookupGenes, base_name='gene')
+router.register(r'laboratories', api_views.LookupLaboratories, base_name='laboratory')
+router.register(r'registries/(?P<registry_code>\w+)/indexes', api_views.LookupIndex, base_name='index')
+router.register(r'registries/(?P<registry_code>\w+)/clinitians', api_views.ListClinitians, base_name='clinitian')
```
rvl: The spelling is *clinician* isn't it? Also the plural of index is indices.

rdrf/rdrf/api_views.py
```diff
@@ -0,0 +1,258 @@
...
+        registry = self._get_registry_by_code(registry_code)
+        if registry not in patient.rdrf_registry.all():
+            self.permission_denied(request, message='Patient not available in requested registry')
+        if request.user.is_superuser:
+            return
+        if registry not in request.user.registry.all():
+            self.permission_denied(request, message='Not allowed to get Patients from this Registry')
+
+        working_groups = patient.working_groups.all()
+        if not any(map(lambda wg: wg in working_groups, request.user.working_groups.all())):
```
rvl: What about `not patient.working_groups.filter(id__in=request.user.working_groups.all()).exists()` instead?

rdrf/rdrf/api_views.py
```diff
@@ -0,0 +1,258 @@
...
+            term = request.GET['term']
+        except KeyError:
+            pass
+            # raise BadRequestError("Required query parameter 'term' not received")
+        registry = Registry.objects.get(code=registry_code)
+
+        if not registry.has_feature('family_linkage'):
+            return Response([])
+
+        working_groups = list(request.user.working_groups.all())
```
rvl: I don't think it needs to be converted to a list to be used as a subquery.

rdrf/rdrf/lookup_views.py
```diff
@@ -185,6 +80,11 @@ def _get_relationships(self):
         return [pair[0] for pair in PatientRelative.RELATIVE_TYPES]
 
 
+# TODO I think that for this one the get will be replaced by Django REST framework view
+# The put that switches contexts should be moved. It isn't a lookup view for sure, but also
+# it doesn't feel "right" to mix it into the REST API because it changes things in the
+# session of a user that is using the web UI not consuming the API
+
```
rvl: Agree, kind of. Session context_id set/get should be separate from the api for getting contexts. Whether it's within the DRF api or in a separate view doesn't really matter because the api is already stateful (login session).

rdrf/registry/patients/models.py
```diff
@@ -84,11 +84,15 @@ def __unicode__(self):
 class PatientManager(models.Manager):
 
     def get_by_registry(self, registry):
-        return self.model.objects.filter(rdrf_registry__in=registry)
+        registries = registry if hasattr(registry, '__iter__') else (registry,)
```
rvl: I don't like such overloading. And the list argument form doesn't seem to be used anyway. What about `def get_by_registry(self, *registries):` ?

rdrf/rdrf/models.py
```diff
@@ -1,6 +1,7 @@
 from django.db import models
 import logging
 from django.core.exceptions import ValidationError
+from rest_framework.reverse import reverse as rest_reverse
```
rvl: Seems to be unused

rvl:
Looks good to me.

rdrf/rdrf/api_urls.py
sztamas: Yeah, it was called clinitian previously and I just copied it over.
Both indices and indexes are acceptable (ex. http://grammarist.com/usage/indexes-indices/) but indexes is maybe more common in US so switching to indices.

rdrf/rdrf/api_views.py
sztamas: Yeah, I like it better.

rdrf/rdrf/api_views.py
sztamas: Yes, I think I just copied that code over.

rdrf/registry/patients/models.py
sztamas: Yes, didn't want to break/find out intent of previous code so kept it backwards compatible.
Changed is as suggested.

rdrf/rdrf/models.py
sztamas: Gone.

sztamas:
Thanks for reviewing it.

rvl:
:metal:

PR #5, opened 2016-05-13
RDR-1256 Registry code is read-only

rvl:


mradochonski:
I like the idea. +1 from me.

PR #6, opened 2016-05-16
Exporter/importer initial version.

sztamas:
Initial version of the exporter for review.
Not all models are exported currently but would like to get some feedback on it.
In the meantime I will work on import then complete the export.

Thanks


rdrf/rdrf/export/model_exporters.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.model.objects.filter(code=self.exporter_context['registry_code'])
+
+
+class RegistryFormExporter(ModelExporter):
+    @property
+    def queryset(self):
+        return self.model.objects.filter(registry__code=self.exporter_context['registry_code'])
+
+
+class AdjudicationDefinitionExporter(ModelExporter):
```
id2359: Adjudication code is dead so no need to export these models

rdrf/rdrf/export/model_exporters.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.model.objects.filter(rdrf_registry__code=self.exporter_context['registry_code'])
+
+
+registry_catalogue.register(models.Registry, RegistryExporter)
+registry_catalogue.register(models.RegistryForm, RegistryFormExporter)
+registry_catalogue.register(models.AdjudicationDefinition, AdjudicationDefinitionExporter)
+registry_catalogue.register(models.Section, SectionExporter)
+registry_catalogue.register(models.CommonDataElement, CommonDataElementExporter)
+
+registry_catalogue.register(patientmodels.Patient, PatientExporter)
```
id2359: Will need exporters for:
QuestionnaireResponse (used by DM1),
Query (reports)
CdePolicy
ConsentSection
ConsentQuestion
ConsentValue
DemographicFields
PatientRelative
EmailTemplate
EmailNotification
EmailNotificationHistory
RDRFContext

rdrf/rdrf/export/definitions.py
```diff
@@ -0,0 +1,97 @@
...
+                    models = (
+                        patientmodels.Patient,
+                )),
+                GroupDefinition(
+                    name='MongoDB Data',
+                    dirname='mongodb_data',
+                    collections = (
+                        'cdes',
+                        'history',
+                        'progress',
```
id2359: there's a registry specific fields collection called "registry_specific_patient_data" also in mongo - used by FH only currently

rvl:
- Looks nice and extensible.
- Could `__main__.py` be removed and the django management command machinery used instead? E.g. command line argument handling.
- Maybe put the date and time in exported filenames
- In `catalogue.py` you could write:
```python
  class ModelExporterCatalogue(Catalogue):
      factory = ModelExporter      
```
  And change `Catalogue.__init__` accordingly.
- Is it possible for `Section` model to have `code` as a unique natural key?
- Is it possible for natural key of `RegistryForm` to be `(registry.code, name)`?
- Are you going to address row ID collisions on import, or not at this stage?

rdrf/rdrf/export/model_exporters.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.model.objects.filter(registry__code=self.exporter_context['registry_code'])
+
+
+class SectionExporter(ModelExporter):
+    @staticmethod
+    def get_sections(registry_code):
+        registry = models.Registry.objects.get(code=registry_code)
+        for form in models.RegistryForm.objects.filter(registry=registry):
+            for section_code in form.get_sections():
+                yield models.Section.objects.get(code=section_code)
```
id2359: I think I did add a property called section_models on RegistryForm to do this  

for section_model in form.section_models:
   yield section_model

rdrf/rdrf/export/model_exporters.py
```diff
@@ -0,0 +1,72 @@
...
+    def queryset(self):
+        return self.get_sections(self.exporter_context['registry_code'])
+
+
+class CommonDataElementExporter(ModelExporter):
+    @property
+    def queryset(self):
+        sections = SectionExporter.get_sections(self.exporter_context['registry_code'])
+        for cde in reduce(lambda res, sec: res.union(sec.cde_models), sections, set()):
+            yield cde
```
id2359: yield cde for cde in set.union(*map(set, [sec.cde_models for sec in sections]) maybe

basic idea on stackoverflow

sztamas: I guess it's objective which one reads better. Some possibilities:

for cde in reduce(lambda res, sec: res.union(sec.cde_models), sections, set()):
for cde in set.union(*map(set, [sec.cde_models for sec in sections]):
for cde in set.union(*map(lambda s: set(s.cde_models), sections)):
for cde in set.union(*[set(s.cde_models) for s in sections]):

Maybe I will go with:

```python
def queryset(self):
        sections = SectionExporter.get_sections(self.exporter_context['registry_code'])
        all_unique_cdes_in_sections = set.union(*[set(s.cde_models) for s in sections])
        for cde in all_unique_cdes_in_sections:
            yield cde
```

rdrf/rdrf/export/model_exporters.py
sztamas: @rvl 
* thanks
* absolutely, but for testing it was easier this way. Plus there is some import there as management command already. We will have to decide what to do about the clash.
* Date and time is written out to the META file. I thought that will be enough.
* Yes, good point. Made the change.
* yes I thinks so. @id2359?
* probably yes. @id2359?
* not addressing it now, but if we will I guess it will have to be on import when we detect them, right?

rdrf/rdrf/export_import/definitions.py
```diff
@@ -0,0 +1,205 @@
...
+            dirname='registry_data',
+            datagroups=(
+                GroupDefinition(
+                    name='Demographic Data',
+                    dirname='demographic_data',
+                    models = (
+                        # Leave RDRFContexts before Patients, so that no
+                        # default contexts are created on import. This works
+                        # because there is no FK from RDRFContext to Patient
+                        'rdrf.RDRFContext',
```
id2359: We'll also have to slot these models in somwhow for 1.5.0: RDRFContext 
ContextFormGroup
ContextFormGroupItem
MongoMigrationDummyModel
CDEFile

rvl: That raises the question, how do FileFields get exported? e.g. `PatientConsent.form` and `CDEFile.item`

rvl:
The code is very cool and I can't find anything really wrong with it.

Here is a list of comments which you are free to implement or not.

rdrf/rdrf/export_import/definitions:83: tuple() -> ()
rdrf/rdrf/export_import/definitions:99: raise Exception -> raise ValueError
rdrf/rdrf/export_import/definitions:182: How do FileFields get exported?

rdrf/rdrf/export_import/exporters.py:44: any further abstraction could obscure the meaning of the procedure
rdrf/rdrf/export_import/exporters.py:76: do it as `set(m.split('.', 1)[0] for m in models)`
rdrf/rdrf/export_import/exporters.py:77: dictionary comprehension is possible

rdrf/rdrf/export_import/main_importers.py:60: this is the same as `json.load(f)`
rdrf/rdrf/export_import/main_importers.py:60: it's also possible to rely on refcounting to close the file. i.e. `self.meta = json.load(open(meta_file))`
rdrf/rdrf/export_import/main_importers.py:191: mongo rollback could be achieved by keeping a list of mongo document ids that were inserted into collections and just deleting them. not sure how easily it would be combined with django rollback.
rdrf/rdrf/export_import/main_importers.py:207: is it worth bringing in python-dateutil dep just for this?

rdrf/rdrf/export_import/importers.py:80: raise Exception -> raise ImporterError. not sure i like the use exceptions for normal control flow.
rdrf/rdrf/export_import/importers.py:121: this is relying on a python misfeature. it would be same length but clearer to just increment actual_object_count after each iteration.

rvl:
Also, can we slip in the PyYAML snippet to format multiline strings properly and replace `\r\n` with `\n`?

id2359:
Speaking of  Mongo rollback - I did that with the same technique Rodney suggested in my script to update FH just recently

rdrf/rdrf/export_import/definitions.py
sztamas: Added ContextFormGroup, ContextFormGroupItem, and CDEFile.
Is MongoMigrationDummyModel really needed? Based on the documentation of the model we shouldn't really export/import it. It isn't the VERSIONS that will get exported it is the rows in the table. The doc doesn't say that the version field should be updated or that it is meaningful in any way.

sztamas: Files aren't exported in this version.
The JIRA issue has a TODO list, files is one item on that list:
https://ccgmurdoch.atlassian.net/browse/RDR-1215

sztamas:
Thanks Rodney!
Inline comments:

rdrf/rdrf/export_import/definitions:83: tuple() -> ()
    I use tuple() because empty tuple is actually (,) not (). I think tuple() is clearer than (,).
rdrf/rdrf/export_import/definitions:99: raise Exception -> raise ValueError
    Done
rdrf/rdrf/export_import/definitions:182: How do FileFields get exported?
    Not done in this version. On the TODO list.
rdrf/rdrf/export_import/exporters.py:44: any further abstraction could obscure the meaning of the procedure
    Yeah, removed TODO comment.
rdrf/rdrf/export_import/exporters.py:76: do it as set(m.split('.', 1)[0] for m in models)
    I made it on 2 lines to document that what I'm extracting is the app_label.
rdrf/rdrf/export_import/exporters.py:77: dictionary comprehension is possible
     Yes, done. I'm still in <2.7 land apparently :)
rdrf/rdrf/export_import/main_importers.py:60: this is the same as json.load(f)
     Yes, done.
rdrf/rdrf/export_import/main_importers.py:60: it's also possible to rely on refcounting to close the file. i.e. self.meta = json.load(open(meta_file))
    Yeah, maybe. What happens on Exceptions? When does the garbage collector kick in? On all implementations we're using? The with construct is idiomatic python code and doesn't raise any of those question.
rdrf/rdrf/export_import/main_importers.py:191: mongo rollback could be achieved by keeping a list of mongo document ids that were inserted into collections and just deleting them. not sure how easily it would be combined with django rollback.
    Yeah, I guess. A bit of poor man's transactions but probably better than nothing. Added a TODO to implement it.
rdrf/rdrf/export_import/main_importers.py:207: is it worth bringing in python-dateutil dep just for this?
    Maybe not, but then maybe other code will need to parse dates as well.
rdrf/rdrf/export_import/importers.py:80: raise Exception -> raise ImporterError. not sure i like the use exceptions for normal control flow.
    Ok. Not sure about the second part. I am raising exceptions when there is something wrong with the file, inconsistency between meta data and data etc.
rdrf/rdrf/export_import/importers.py:121: this is relying on a python misfeature. it would be same length but clearer to just increment actual_object_count after each iteration.
    Yes, changed it.

rvl:
Sweet, cool, thanks.

1. Try this in python shell: `repr(tuple())`.
2. Python is ref counted not garbage collected. I'm pretty sure opening a `with` block is just the same as using the function block scope (because the `with` block is the entire function).

:+1: :ok_hand: :ship:

sztamas:
1. ah ok, you're right. I confused it with a single element tuple. ('x') != ('x',)
2. Yes, that is true for cpython. What happens when an exception is thrown if you don't use the with block. Is it guaranteed that the file will be closed?

grahame:
Python's not defined to be reference counted, that's just cPython. Pypy doesn't do refcounting, it does GC, so 'with' is better for files and other resources which you need to ensure are cleaned up at the end of a scope no matter what happens. (http://pypy.org/compat.html)

PR #8, opened 2016-05-17
CDE Value History Charts and Revert

rvl:
Click on little chart icon next to CDE field label to see previous snapshot values.

Numeric fields are plotted with a time series.

Non-complex fields can have their values reverted.

Field history view has the same permissions as form view, however it seems odd to me.

It might be good to switch on django timezone support and store timestamps in ISO format.

I updated Chart.js to version 2.1.0 because it adds supports for dates on the x-axis.

03-cde-value-history-chart


aahunter:
FYI this is https://ccgmurdoch.atlassian.net/browse/RDR-1266 not 1236

PR #9, opened 2016-05-17
Django file storage

rvl:
## RDR-1265

This change uses Django to store newly uploaded files.

The code allows files to be accessed from either GridFS or Django. So no migration step is immediately necessary.

I looked into setting up a GridFS Django storages backend but it's complicated because the Mongo DB changes depending on the registry – so didn't bother.

The `CDEFile` model links a file to the CDE, registry, and (optionally) form and section.

The GridFS → Django storage copy script is a django management command. It will copy each file into Django storage, update the CDE, delete the GridFS file. It has a `--dry-run` option for testing.

## RDR-1235

Multiple files can be stored in a single file CDE if the `allow_multiple` field is true.

The user interface looks like this:

rdr-1235-multifile-cde

I also did a few cleanups to try and make the code shorter and easier to follow.


rdrf/rdrf/filestorage.py
```diff
@@ -1,39 +1,93 @@
...
-        logger.debug("gridfs_ref_dict = %s" % gridfs_ref_dict)
-        return gridfs_ref_dict
-
-    def delete(self, gridfs_file_id):
-        if self.gridfs_filestore.exists(gridfs_file_id):
-            self.gridfs_filestore.delete(gridfs_file_id)
-            logger.info("GridFS API: deleted %s OK" % gridfs_file_id)
-        else:
-            logger.info("%s did not exist in filestore")
+class FileStoreUtil(object):
```
sztamas: Polymorphism could be used here to get rid of the @staticmethods and the if gridfs do-gridfs-stuff else do-django-file-stuff constructs?

rdrf/rdrf/management/commands/migrate_gridfs_django.py
```diff
@@ -0,0 +1,183 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("--dry-run", action="store_true",
+                            help="Don't do anything, just log.")
+        parser.add_argument("registry", nargs="*")
+
+    def handle(self, *args, **options):
+        all_codes = Registry.objects.values_list("code", flat=True)
+        for registry_code in options["registry"]:
+            if registry_code not in all_codes:
+                logger.warning("Registry \"%s\" does not exist." % registry_code)
```
sztamas: 'Registry "%s" does not exist.' % registry_code
reads better IMHO. PEP-8 also says so.

rdrf/rdrf/management/commands/migrate_gridfs_django.py
```diff
@@ -0,0 +1,183 @@
...
+                   item=File(file_obj),
+                   filename=filename or "")
+
+
+def calc_md5(file_obj):
+    h = hashlib.new("md5")
+    while True:
+        buf = file_obj.read(16*1024)
+        if not buf:
+            break
```
sztamas: Second version of iter(), the one with sentinel values, reads better IMHO.
No while True, no break on end value.

rdrf/rdrf/filestorage.py
rvl: I can't see how polymorphism would make things better because the id type needs to be tested for each `file_ref` value.

Actually these methods shouldn't even be in a class.

sztamas: A static method creates the right instance based on the id type check, then you just call the method (delete(), etc. on the instance).

Or maybe the methods shouldn't be in a class as you say.

Generally, if I see a static/class with branches based on type inside the solution is usually polymorphism. Just made a note, you'll know if it applies here or not.

rvl:
Thanks @sztamas for the review I implemented your suggestions. Object method polymorphism on file refs wasn't feasible. I tried switching to single quotes but that meant that apostrophes needed to be escaped.

There will be more commits coming on this branch for multiple file upload, which is proving to be quite a pain...

sztamas:
I didn't mean switching to single quotes for all strings. You should pick
one style of quoting (single or double) for all your strings. Then for
strings that contain the quote character of your choice, instead of
escaping it you would  switch to the other style of quoting.

rdrf/rdrf/dynamic_data.py
```diff
@@ -30,79 +24,102 @@ class KeyValueMissing(Exception):
     pass
 
 
-class MultisectionGridFSFileHandler(object):
-    def __init__(self,
-                 gridfs_filestore,
```
id2359: glad this is being culled - it was hideous:)

rdrf/rdrf/dynamic_data.py
```diff
@@ -30,79 +24,102 @@ class KeyValueMissing(Exception):
...
+    for f, form in enumerate(doc.get("forms") or []):
+        if (form_name is None or
+            (form.get("name") == form_name and formp(form, f))):
+            for s, section in enumerate(form.get("sections") or []):
+                if (section_code is None or
+                    (section.get("code") == section_code and
+                     sectionp(section, s))):
+                    yield section
+
+def find_cdes(doc, form_name=None, section_code=None, cde_code=None,
```
id2359: good idea  - I was going to do something similar in utils but good here.   Also I wonder if we could annotate these generators with contextlib.contextmanager

rdrf/rdrf/filestorage.py
```diff
@@ -1,39 +1,88 @@
-# Attempt to clean up the interaction with GridFS
+"""
+Stored files are referenced in CDEs as a dict like this:
+    file_ref = {
+        "gridfs_file_id": ObjectId(...), # gridfs is no longer used
+        "django_file_id": 1,             # reference to CDEFile.id
+        "filename": "goatse.jpg",
```
id2359: :)

id2359:
Noticed that when a file cde is in a multisection ( dd.yaml is a good test registry for this - the diagnosis form has an instance )

Then  if a file is uploaded into one of the added "sections" of the multisection and then subsequently "cleared" in the GUI - a run time error occurs:

File "/app/rdrf/rdrf/dynamic_data.py", line 59, in find_cdes
(cde.get("code") == cde_code and
AttributeError: 'list' object has no attribute 'get'

I think this is because within the mongo record the "cdes" key is a list of cde dicts for a normal section
but for a multisection (allow_multiple=True) the key "cdes" is a list of the sections comprising the multisection ( themselves lists of cde dicts) ... so the implementation of find_cdes needs to be aware of this
I initially was going to have a separate items key only for multisections but we rejected it

rdrf/rdrf/dynamic_data.py
```diff
@@ -30,79 +24,102 @@ class KeyValueMissing(Exception):
...
+    """
+    cdep = cdep or (lambda c, i: True)
 
-        # bookkeeping of gridfs file ids
-        self.existing_file_ids = set([])
+    sections = find_sections(doc, form_name, section_code, formp, sectionp)
+    for section in sections:
+        for c, cde in enumerate(section.get("cdes") or []):
+            if (cde_code is None or
+                (cde.get("code") == cde_code and
```
id2359: This fails for multisections - the dict looks like this for multisections:
{
"code" : "ImAMultisectionCode",
"allow_multiple": True,
"cdes": [ list of "items"  in the multisection ...  where each item is just like a non-multiple section dict ]
ie if a MS was defined as cdeA cdeB then
cdes list might look like [ <item1>, <item2> ]
item1 = [ { first cde dict for A} , {first cde dict for B}]

item2 = [ {second cde dict for A} , {second cde dict for B}

rdrf/rdrf/filestorage.py
```diff
@@ -1,39 +1,76 @@
-# Attempt to clean up the interaction with GridFS
 import logging
+import re
+import gridfs
+from bson.objectid import ObjectId
+from .models import Registry, CDEFile
```
id2359: CDEFile model has link to section but what  if the linked section is a "multisection?

Do we need a nullable index field also ( ie null in the case where the section is a "normal" (non-)multisection 0-numitems-1 otherwise 

After looking I can see there's no need as it's deducible from the mongo record ) ( as is the fact that it "belongs" to a given patient  - We can report on mongo cdes including file cdes using the explorer so this shouldn't matter ( although some queries would be easier eg find all patients who haven't uploaded a catscan file ))

rvl:
OK `find_cdes()` and multisection saving has been fixed.

rdrf/rdrf/filestorage.py
rvl: The multisection index field would be tricky because they might need to be updated if a section is deleted.

I was thinking the same thing, that the `CDEFile` is implicitly linked to the mongo document. They might come in handy. I would prefer to keep the references for now and ignore them when necessary.

PR #12, opened 2016-05-31
Init management command

sztamas:
```
$ django-admin init --list

Possible datasets are:
  - base
      The minimal data needed for proper functioning.
  - DEV
      Collection of datasets to be used by developers.
  - groups
      Initial groups for different roles.
  - users
      Initial users with different roles.
  - sample_registry
      Sample registry.
  - sample_laboratories
      Sample laboratories from AU and NZ...
  - genes
      A large list of genes.
  - genes_smaller_sample
      A small sample of genes.
  - iprestrict_permissive
      Disable iprestriction completely.
```

  - not loading data from migrations anymore
  - removed the old fixtures, so they don't confuse anyone
  - instead of loading dev fixtures from ``docker-entrypoint.sh`` we're calling ``init DEV``
  - I used genes_smaller_sample for DEV, that is around 80 rows of data not almost 40,000 as the main genes dataset

I could have spent more time on setting up more data to be loaded by DEV, but maybe we will do that by doing an import instead.


rvl:
Looks fine to me.

 - Why is the genes dataset split up? It's not that big.
 - A large number of datasets would confuse people. Some of them don't seem independently useful.
 - I would remove people's office phone numbers from the sample data.

sztamas:
Thanks for reviewing it.  

* it takes quite a bit of time to export/import/init all those genes. I don't think it is worth having all of them in dev.
  * I removed a few of them. Currently only DEV is used. I think we will have only a few like DEV for different environments combining others together. The list is now:

```
Possible datasets are:
  - DEV
      Collection of datasets to be used by developers.
  - groups
      Initial groups for different roles.
  - users
      Initial users with different roles.
  - genes
      A large list of genes.
  - iprestrict_permissive
      Disable iprestriction completely.
```

  * those office phone numbers were in the fixtures, I just moved them over. I'm not sure if they are real or not but I will modify them a bit.

aahunter:
The subset list of genes for dev is worthwhile. Waiting for that migration is a pain.

rvl:
I'm not necessarily sugggesting to remove the smaller gene dataset. Just to roll it into DEV because it's not independently useful.

I don't think migrate+init should be run every time the server starts, but that's another subject.

What about making genes import conditional on already existing data, i.e.
```python
if models.Gene.objects.count() < 30000:
    print("Loading genetic data, this will take 3-4 minutes")
    # ...
else:
    print("Genetic data already loaded")
```

Also for next time, instead of making a new JSON file, maybe just every ~30th item could be loaded from the existing file.

There is also an empty file called "gene" in the patch which can be removed.

sztamas:
Yes it is rolled into DEV now.

I agree that migrate+init shouldn't run each time we start the server.
collectstatic shouldn't run, uwsgi shouldn't be started up etc.

if you make it conditional you will still have to wait each time you start
from scratch.
For example when testing import many times and you want to do it in a new
installation of RDRF with an empty DB each time.

Yes, maybe. This way it is more deterministic but there is an extra file.
Pros and cons not sure it matters that much.

Thanks, removed gene.

rvl:
:ok_hand:

rvl:
I don't see any reason why this couldn't be merged.

PR #13, opened 2016-06-16
Add a script for running registry yaml csv test cases

rvl:
For an example usage, see: https://bitbucket.org/ccgmurdoch/rdrd/src/next_release

### DIFF
--- a/scripts/run_yaml_cde_calculation_tests.py
+++ b/scripts/run_yaml_cde_calculation_tests.py
@@ -23,8 +23,11 @@ def main():
     parser.add_argument("test_csvs", metavar="CSV", nargs="+",
                         type=argparse.FileType("r"),
                         help="CSV files containing test definitions")
-    parser.add_argument("--outfile", metavar="FILE", type=argparse.FileType("w"),
+    parser.add_argument("--outfile", metavar="FILE", type=argparse.FileType("wb"),
                         help="Output file")
+    parser.add_argument("--verbose", "-v", action="count",
+                        help="More info for debugging tests")
+
     args = parser.parse_args()
 
     if not args.outfile: args.outfile = sys.stdout
@@ -33,7 +36,7 @@ def main():
 
     success = True
     for infile in args.test_csvs:
-        success = run_tests(registry, infile, args.outfile) and success
+        success = run_tests(registry, infile, args) and success
     return 0 if success else 1
 
 Registry = namedtuple("Registry", ("names", "calculations"))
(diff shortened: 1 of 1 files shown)


scripts/run_yaml_cde_calculation_tests.py
```diff
@@ -0,0 +1,140 @@
...
+    params = ((code, val) for code, val in cols.items() if code != "check")
+    return TestCase(filename, num, check_code, dict(params))
+
+def load_adsafe_js():
+    adsafe_path = "../rdrf/rdrf/static/js/vendor/adsafe-min.js"
+    return io.open(os.path.join(os.path.dirname(__file__), adsafe_path)).read()
+
+def run_test(registry, test):
+    context = dict((code, val) for code, val in test.params.items() if code in registry.names)
+    patient = dict((code, val) for code, val in test.params.items() if code not in registry.names)
```
id2359: I guess that's slightly brittle if someone did happen to create a cde code with the same name as a patient attribute - unlikely though

scripts/run_yaml_cde_calculation_tests.py
```diff
@@ -0,0 +1,140 @@
...
+        elif res.expected == res.actual:
+            num_success += 1
+        else:
+            print_failure(registry, res, row, outfile)
+
+    return num_tests == num_success
+
+def setup_test(cols, num, filename):
+    check_code = cols["check"]
+    params = ((code, val) for code, val in cols.items() if code != "check")
```
id2359: Just querying your use  of "check_code": Currently the JS calcs are wired up so that the special field context.result holds the final result of the calculation - this gets passed on to the calculated field cde widget on the page - so if the calculated field code was cdeBMI , say, context.cdeBMI is not updated in the calculation body, but context.result is. The context is pre-populated on the page with the input cde values ( using their codes as attributes) context.cdeWeight or context.cdeHeight say -  "

scripts/run_yaml_cde_calculation_tests.py
```diff
@@ -0,0 +1,140 @@
...
+        var document = {};
+        %s
+
+        var patient = %s;
+        var context = %s;
+        var RDRF = ADSAFE;
+
+        %s
+
+        console.log(context.result);
```
id2359: ok so you are getting context,.result - cool:)

scripts/run_yaml_cde_calculation_tests.py
id2359: ok forget that I see you're doing now

scripts/run_yaml_cde_calculation_tests.py
```diff
@@ -0,0 +1,140 @@
...
+    """ % (load_adsafe_js(), json.dumps(patient), json.dumps(context), script)
+
+    success, output = exec_script(script)
+
+    if success:
+        lines = [o.strip() for o in output.split("\n")]
+        non_empty = filter(bool, lines)
+        context_result = non_empty[-1] if non_empty else None
```
id2359: My only concern would be we're testing  the result of writing context.result to a log file not context.result directly - am I right?

rvl:
> My only concern would be we're testing the result of writing context.result to a log file not context.result directly - am I right?

Yes that's right, the testing of `context.result` is done outside of javascript. Another possibility would be writing the test output line in json format.

I think we might need to write some more test cases then see if anything needs tweaking.

rvl:
I wrote some test cases for [CDEfhDutchLipidClinicNetwork](https://bitbucket.org/ccgmurdoch/rdrd/src/next_release/tests/fh/dutch-lipid.csv).

The script is updated to allow for more logging/debug info, and unicode characters in the CSV.

PR #15, opened 2016-06-27
RDR-1348 Minor Python dependency upgrades

aahunter:
Minor dep upgrades, will do pymongo on its own. Tests pass.


id2359:
approved

PR #16, opened 2016-06-28
RDR-1350 pymongo 2.9.3

aahunter:
Upgrade path for mongo.

https://api.mongodb.com/python/2.9/changelog.html
https://api.mongodb.com/python/2.9/migrate-to-pymongo3.html

Checked the migration path, none of the deprecated api in the code base.


id2359:
approved

PR #18, opened 2016-07-01
Rdr 1357 deprecate django secure

aahunter:
django-secure was deprecated in Django 1.8 and the functionality is now core Django.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -84,6 +84,12 @@ function selenium_defaults {
 }
 
 
+function _django_check_deploy {
+    echo "running migrate"
+    django-admin.py check --deploy --settings=${DJANGO_SETTINGS_MODULE} 2>&1 | tee /data/uwsgi-check.log
+}
+
+
 function _django_migrate {
     echo "running migrate"
     django-admin.py migrate --noinput --settings=${DJANGO_SETTINGS_MODULE} 2>&1 | tee /data/uwsgi-migrate.log
@@ -147,6 +153,7 @@ if [ "$1" = 'uwsgi' ]; then
 
     _django_collectstatic
     _django_migrate
+    _django_check_deploy
 
     exec uwsgi --die-on-term --ini ${UWSGI_OPTS}
 fi
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -85,7 +85,7 @@ function selenium_defaults {
 
 
 function _django_check_deploy {
-    echo "running migrate"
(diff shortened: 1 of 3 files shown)


rvl:
Looks fine to me. :star:

* I would favour simply enabling the clickjacking protection and not even making it an option.
* Likewise for many other settings, do they actually ever need to be changed?
* HSTS options should be removed from Django settings. These are better handled in the reverse proxy config, alongside the ssl certificate configuration.
* On a related topic, can we consider not fetching jquery, etc from 3rd party CDNs?

aahunter:
Thanks Rodney.

The only thing I can think of is it might be useful to toggle them in dev to isolate/test potential issues. Perhaps that is a long bow though.

For HSTS perhaps a comment in the config that they should be handled upstream will suffice.

Agree with the off-topic comment about CDN usage. A workable alternative might be the integrity attribute. 

https://stackoverflow.com/questions/32039568/what-are-the-integrity-and-crossorigin-attribute

grahame:
+1 to dealing with HSTS in the reverse proxy config. It's much easier and it can be annoying having to trim the header off in the reverse proxy to add a correct version. User-Agents get upset when they see two HSTS headers.

aahunter:
moved cdn comment to #20

aahunter:
Comments addressed. If there are no more comments, will merge. @rvl @id2359

id2359:
Looks fine to me, I can't imagine any reason we'd allow our views to be
framed.

On Tue, Jul 5, 2016 at 11:15 AM, Adam Hunter <notifications@github.com>
wrote:

> Comments addressed. If there are no more comments, will merge. @rvl
> <https://github.com/rvl> @id2359 <https://github.com/id2359>
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/muccg/rdrf/pull/18#issuecomment-230379708>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABHCd3HHKlJkOrx3CNQVFQ7gsO_Yx9TZks5qScw5gaJpZM4JC1eJ>
> .
>



-- 

*********************************************************

Lee Render
Software Engineer
Centre for Comparative Genomics
Murdoch University
Perth, Western Australia, 6150ccg.murdoch.edu.au

Email: *lrender@ccg.murdoch.edu.au <lrender@ccg.murdoch.edu.au>*

*Phone number:* +61-8-9360 6088
*Fax number:* +61-8-9360 7238
*********************************************************

PR #19, opened 2016-07-01
RDR 1358 remove suits

aahunter:
suits isn't installed but lingering references remain, removed references


id2359:
sounds like a good idea - just fyi we do still use the admin interface though

aahunter:
@id2359  Can you point me to somewhere these changes might be a problem?

id2359:
Looks like Maciej has cleaned up pages linked from Admin Pages but you could try http://localhost:8000/admin/groups/customuser/ for a test - I can't find suit referenced on the pages I've tested locallly just now so should be ok.

id2359:
only the error pages were using ( which you fixed) - so ok

PR #20, opened 2016-07-05
RDR 1359 no cdn

aahunter:
PR to remove CDN usage from JS/CSS.

There are a number of related issues - the state of the JS/CSS usage in general. I've only focused on removing the CDNs. Where it didn't break things I combined JS library versions. For example, down from 4 (!) versions of bootstrap to 2. 

The commits are separated into commits with deps and a single final commit with all the changes to templates. The template changes are done with find/xargs/sed.


aahunter:
The other comments are valid, but for this ticket just going to focus on removing the CDN usage.

PR #22, opened 2016-07-07
RDR-1362 add root logger

aahunter:
Adds a root logger.

### DIFF
--- a/rdrf/rdrf/settings.py
+++ b/rdrf/rdrf/settings.py
@@ -357,8 +357,10 @@ LOGGING = {
         },
     },
     'loggers': {
-        'django': {
+        '': {
             'handlers': ['console', 'file'],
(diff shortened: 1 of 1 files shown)


id2359:
good to go

rdrf/rdrf/settings.py
```diff
@@ -357,8 +357,10 @@
         },
     },
     'loggers': {
-        'django': {
+        '': {
             'handlers': ['console', 'file'],
+            'level': 'DEBUG' if DEBUG else 'INFO',
+            'propagate': True
         },
         'django.request': {
```
sztamas: Can we make this '' instead of 'django.request' to mail all CRITICAL logs to admins?

rvl:
What does the `'DEBUG' if DEBUG else 'INFO'` conditional achieve, that isn't already handled by the `level` and `filters` options on the log handlers?

PR #24, opened 2016-07-07
Upgrade to django-iprestrict 0.4.3

sztamas:
@aahunter Can you please update puppet with the new env variables?


id2359:
approved

PR #26, opened 2016-07-09
RDRF-1373: Patient Listing Page - handle edge cases better

sztamas:
Shows an alert message if:

  * the user has no registry assigned to him
  * the user has no permissions to see any of the columns in the table

instead of failing in JS and leaving the page in "Loading..." state.

### DIFF
--- a/rdrf/rdrf/form_view.py
+++ b/rdrf/rdrf/form_view.py
@@ -2695,8 +2695,10 @@ class ContextsListingView(LoginRequiredMixin, View):
 
         context["columns"] = columns
 
+        template = 'rdrf_cdes/contexts_no_registries.html' if len(registries) == 0 else 'rdrf_cdes/contexts.html'
+
         return render_to_response(
-            'rdrf_cdes/contexts.html',
+            template,
             context,
             context_instance=RequestContext(request))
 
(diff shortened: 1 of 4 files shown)


rdrf/rdrf/templates/rdrf_cdes/contexts.html
```diff
@@ -153,22 +158,34 @@
...
-        <span id="loading_text"><i>Loading...</i></span>
-    </p>
-
-    <table id="context_list_table" class="table table-striped table-hover">
-        <thead>
-            <tr>
-                {% for column in columns %}
-                    <th>{{ column.label|safe }}</th>
+    {% if not registries %}
+        <div class="alert alert-danger">
```
id2359: seems ok, if there are no registries , what happens to the JS that hooks up  the ( non existent) select with id registry_options?

id2359:
If the commented issue fails silently, then ok , approved

rdrf/rdrf/form_view.py
```diff
@@ -2695,8 +2695,10 @@ def get(self, request):
 
         context["columns"] = columns
 
+        template = 'rdrf_cdes/contexts_no_registries.html' if len(registries) == 0 else 'rdrf_cdes/contexts.html'
```
id2359: Yeah this is nice

id2359:
This is definitely cleaner - cool. Approved

PR #27, opened 2016-07-11
Relax User validation, move permissions and grid config out of settings

sztamas:


rdrf/registry/groups/admin_forms.py
```diff
@@ -5,26 +5,26 @@
...
+        working_group_models = self.cleaned_data.get("working_groups", [])
 
-        if "working_groups" in self.cleaned_data:
-            working_group_models = [working_group_model for working_group_model
-                                    in self.cleaned_data["working_groups"].all()]
-        else:
-            working_group_models = []
+        if len(registry_models) == 0 and len(working_group_models) == 0:
+            # Registries and working groups not selected. Don't check for consistency.
+            return self.cleaned_data
```
id2359: Seems ok - is there any where in the code where len(registry_models) just assumed > 0 or len(working_groups) >0

id2359:
Approved

PR #31, opened 2016-07-23
Lettuce scenarios for basic registry

sztamas:


aahunter:
This looks pretty good, I would be keen to see this on next_release.

My only comment would be: Can we factor out the coupling between the docker-entrypoint and the tests. This will get cumbersome as we add more scenarios. Perhaps move  _django_iprestrict_permissive_fixtures and _import_dd_registry into the steps file and make runserverlettuce redundant.

sztamas:
Yes, we can ... sort of.

The main difference between the runserver and runserver_lettuce is the data that is loaded.
The DEV init data used in the runserver includes the iprestrict permissive stuff and also sets up the "Sample registry". We want the iprestrict permissive stuff, but don't want the "Sample registry" set up.
So, even if I move the registry import into the lettuce setup there will still be a difference, we don't want the "Sample registry" to be set up when running lettuce. The import currently expects an empty database, so if we set up "Sample registry" and then import DD on top of it we will have to --force the import through. Not clean and there no scenario that needs "Sample registry".

So maybe move the dd import to lettuce as you say, but have an option in runserver to optionally bypass setting up "Sample Registry". I have a feeling that setting up "Sample registry" will go away anyways. Instead we will have an import of a more useful "dev" registry, that will also have patient data, both demographic and CDEs.

What do you think?

aahunter:
Understood and agree. 

As long as we don't have to keep coming back to the entrypoint everytime we add a new scenario/import etc.

aahunter:
please update this with next_release, the latest lettuce changes have landed there

sztamas:
I talked to Lee about it and he started his branch (fhtests) with all the lettuce tests from here.
Since I haven't added more tests since then it is easier to just drop this PR.
fhtests is already merged into next_release and it has all the tests from here.

PR #42, opened 2016-08-22
Lettuce teardown and vnc

aahunter:
--no-teardown to leave database and browser intact after test run.
Suggestion from Tamas, use debug images to support vnc. vnc to localhost:5900, pword is secret.


id2359:
What's the use case for LETTUCE_DISABLE_TEARDOWN being true?

id2359:
approved

PR #43, opened 2016-08-22
Lettuce: running without alert confirmation hacks

sztamas:
Fixes:

https://ccgmurdoch.atlassian.net/browse/RDR-1433
https://ccgmurdoch.atlassian.net/projects/RDR/issues/RDR-1434

  * changed ``fh.zip`` so the calculated fields in it return empty string when they were returning ``NaN``
  * we don't do the hack of automatically accepting possible alerts after each step
  * modified the changed detection algorithm in the forms to not take readonly field into account. Calculated fields (which are readonly) always are incorrectly detected as changed otherwise.


id2359:
approved

PR #231, opened 2016-09-06
enable lazy-apps uwsgi option

grahame:
For an explanation of what lazy-apps does, see:
https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html#preforking-vs-lazy-apps-vs-lazy

By default, uwsgi loads the state of the app, including all libraries,
and then forks. This causes problems for some libraries, such as
SQLAlchemy:
https://stackoverflow.com/questions/22752521/uwsgi-flask-sqlalchemy-and-postgres-ssl-error-decryption-failed-or-bad-reco

To play it safe, we should enable lazy-apps.

### DIFF
--- a/uwsgi/docker.ini
+++ b/uwsgi/docker.ini
@@ -5,4 +5,5 @@
 emperor      = /app/uwsgi/vassals
 touch-reload = /app/uwsgi/docker.ini
 vacuum       = True
+lazy-apps    = True
 env          = DJANGO_SETTINGS_MODULE=rdrf.settings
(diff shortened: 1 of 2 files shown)


uwsgi/docker.ini
```diff
@@ -5,4 +5,5 @@
 emperor      = /app/uwsgi/vassals
 touch-reload = /app/uwsgi/docker.ini
 vacuum       = True
+lazy-apps    = True
```
aahunter: Perhaps remove from this file.

uwsgi/docker.ini
```diff
@@ -5,4 +5,5 @@
...
 env          = DJANGO_SETTINGS_MODULE=rdrf.settings
```
aahunter: Unrelated to PR, but this env settings should not be needed.

grahame:
@aahunter thanks, I've done those.

aahunter:
:+1:

PR #233, opened 2016-09-08
form_view: Fix missing file cde download link in post view

rvl:


rdrf/rdrf/file_upload.py
```diff
@@ -62,9 +66,38 @@ def wrap(value, key):
...
     return wrap(data, None)
+
+
+def merge_gridfs_data_for_form(registry, form_data, dyn_patient):
+    """
+    Merges existing patient mongo from dyn_patient into form_data, if
+    it's missing.
+    Because form_data is cleaned_data from a Django form, file uploads
+    are missing. In order to display the currently existing files,
+    they need to be copied over from dyn_patient.
```
id2359: Yes I thought this was the problem I caused after changing the form section validation to all or nothing

id2359:
Maciej has tested locally to confirm  - approved

id2359:
I've merged it Rodney, thanks

id2359:
We've had some issues testing on staging, Rodney: Here are my findings:
There are two issues:
a) The download link appears ( which is good) upon upload but doesn't have correct target ( it just reloads the main form , subsequent click works ( ie in the get handler)

 b) A regression with multisections. Again it's in the post-request handler's re-rendering of the sections after upload; the get view is ok , but for file cdes within a multisection, the first download link does not appear if a second file  in another section is uploaded ; When the form is viewed in a subsequent  get request, both links appear and allow files to download. which indicates the problem is in the form initialisation in the response of the post handler .

rvl:
a) Yeah this is "works as designed." :smiling_imp: It would be fairly difficult to fix because the form generation code runs before the CDE saving code. So the database ID of the newly uploaded file is unknown, hence a link can't be generated. Clicking the link to refresh the page seems innocuous enough. Besides, why would a user want to download the file they just uploaded?

b) OK, there are code branches for two cases -- single section, multiple sections, and I missed the second one. Fixed in #238.

PR #242, opened 2016-09-20
Create utility script to re-export test zipfiles

### LINKED ISSUE #240: Create utility script to re-export test zipfiles
The lettuce tests are using exported registries to set up data needed for scenarios.

However, when the schema of our apps change, the importer detects the schema change and refuses to import the zipfiles.

Write a script that iterates through all the lettuce zipfiles, imports them with `--force` then re-exports them.

sztamas:
@aahunter can you please have a look?

Closes #240


lib.sh
```diff
@@ -445,6 +445,30 @@ dev_lettuce() {
...
+  done
+}
+
+reexport_test_zip() {
+    info 'reexport test zips'
+    _start_test_stack --force-recreate -d
+
+    set -x
+    set +e
+    docker-compose --project-name ${PROJECT_NAME} -f docker-compose-unittests.yml run testhost /app/docker-entrypoint.sh /app/reexport_zip.sh $1
```
aahunter: Should this be an exec rather than a run given there is a complete stack already running?
If it has to be a run, then perhaps add --rm

reexport_zip.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
aahunter: If we need this shell snippet then move to ./scripts/
Alternatively the snippet could be absorbed in to the entrypoint and deprecated.

lib.sh
```diff
@@ -445,6 +445,30 @@ dev_lettuce() {
     exit $rval
 }
 
+reexport_test_zips() {
```
aahunter: Avoid anything project specific to lib.sh

sztamas: ok. Should I copy usage to develop.sh too?

PR #252, opened 2016-09-27
importer: add ConsentSection.information_text

### LINKED ISSUE #250: Yaml export/import - 'information text' in 'Consent section' is not retained

rvl:
Seems to have been accidentally missed out. Fixes #250.

### DIFF
--- a/rdrf/rdrf/importer.py
+++ b/rdrf/rdrf/importer.py
@@ -708,13 +708,11 @@ class Importer(object):
             for section_dict in self.data["consent_sections"]:
                 code = section_dict["code"]
                 section_label = section_dict["section_label"]
-                information_link = section_dict["information_link"]
-
                 section_model, created = ConsentSection.objects.get_or_create(
                     code=code, registry=registry, defaults={'section_label': section_label})
-                if not created:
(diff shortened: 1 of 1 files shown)


aahunter:
@id2359 can you review please?

rdrf/rdrf/importer.py
```diff
@@ -708,13 +708,11 @@ def _create_consent_sections(self, registry):
...
                 code = section_dict["code"]
                 section_label = section_dict["section_label"]
-                information_link = section_dict["information_link"]
-
                 section_model, created = ConsentSection.objects.get_or_create(
                     code=code, registry=registry, defaults={'section_label': section_label})
-                if not created:
-                    section_model.section_label = section_label
-                section_model.information_link = information_link
+                section_model.section_label = section_label
```
id2359: Might be better to add a conditional check 

if "information_link" in section_dict:
     then set 

as earlier exported yaml files may not contain this key/value pair 

I've done this before as we'e introduced new model changes into the registry - the earlier exported yamls still need to imported

rvl:
Good point, is the fix in a6cbae9 ok?

PR #255, opened 2016-09-27
patientlisting: Make it faster

rvl:
Investigating a report that the patient listing was slow.

1. Sorting by working group, etc. was throwing an exception, so fixed that.
2. Fixed the way in which diagnosis progress, etc are sorted.
3. Dates were formatted backwards in the patient list, fixed that.
4. Added caching of url reverse() calls.
5. Added caching of the user permissions/forms checking
6. Refactored code so that column functions are together in the same classes.

Two things are still slow:
1. Sorting on columns which aren't table column(s). It needs to load in the entire patient list and sort it before paginating.
2. Looking up dynamic data to get completion status.

To fix these, it would probably be necessary to redesign it. Some caching table could be added but that would further obscure the code.


aahunter:
@id2359 Can you review please?

rvl:
Regarding the two things which are still slow:

1. Sorting columns. By default it sorts by name which is fast. Only if the user tries to sort by e.g. working group, will it take the slow path. Even then, with the number of records we have, it's not too bad. If the records become too many, then we can switch off sorting in the jQuery datatable.
2. Completion status lookup. Lee, if you're refactoring/simplifying the dynamic data code this week, then it might open up possibilities for easier caching. But this can be examined later.

id2359:
Might need to investigate if this change is compatible across all our registries -Maciej will check permissions code

mradochonski:
I am happy with the changes.

PR #260, opened 2016-09-30
exporter: Skip default values in registry yaml dump

rvl:
This makes it easier to read CDEs. If you squint, it almost looks like a DSL for defining registries.

I haven't tested importing of this yaml, so don't merge yet.

### DIFF
--- a/rdrf/rdrf/exporter.py
+++ b/rdrf/rdrf/exporter.py
@@ -4,6 +4,7 @@ import yaml
 import json
 from operator import attrgetter
 from django.forms.models import model_to_dict
+from django.db.models import fields
 from rdrf import VERSION
 import datetime
 from rdrf.models import AdjudicationDefinition, DemographicFields, RegistryForm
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/exporter.py
```diff
@@ -499,6 +505,13 @@ def _get_context_form_groups(self):
             data.append(cfg_dict)
         return data
 
+def cull_defaults(Model, d):
+    for field in Model._meta.fields:
+        if (field.name in d and
+            (d[field.name] == field.default or
+             (not d[field.name] and field.default == fields.NOT_PROVIDED))):
```
rvl: It was because CharFields have an implicit default of `""`.

I think I should replace `not d[field.name]` with `d[field.name] in (None, "")`.

Anyway, this is the least urgent PR out of the list. Main point of this was to illustrate what the yaml looks like without redundant noise.

id2359:
We've had some discussion around this and think it's probably better to keep including defaults despite the reduced readability ( just a trade off).
Keeping defaults 
a) is more explicit.
b) is more robust if ( for whatever reason) defaults change in Django or our app.

Issue #273, opened 2016-10-11
Modules drop down not appearing on next_release for fh import


id2359:
Tested dm1_au yaml which works ( shows modules dropdown in patient listing)

FH yaml is not showing modules drop down though

id2359:
Corrected the check for context form groups on a registry in the context menu column class

PR #275, opened 2016-10-11
Patients listing: contexts and modules dropdown #273

rvl:
Is this logic right?

```
registry supports contexts ⇒ disable modules dropdown in patients listing
```


id2359:
Fixed 1.7.x

id2359:
FH is a special case because it (alone) makes use of "Context form groups"

(cf https://github.com/muccg/rdrf/commit/332e3764768c22f9d62e16325c0167320642342d)

PR #310, opened 2016-11-02
Explorer fixes

rvl:
A fix for the unicode decode error of muccg/rdrf-ccg#848.

I was unable to work out how to actually make it produce the excel report, so the fix is not tested.

### DIFF
--- a/rdrf/explorer/views.py
+++ b/rdrf/explorer/views.py
@@ -1,15 +1,19 @@
-from django.http import HttpResponse
-from django.http import HttpResponseRedirect
-from django.shortcuts import render, redirect
-from django.views.generic.base import View
-from django.core.urlresolvers import reverse
-from django.core.exceptions import PermissionDenied
+from datetime import datetime
+from itertools import product
+import logging
+import re
+from tempfile import NamedTemporaryFile
+from bson.json_util import dumps
 
 from django.contrib.auth.decorators import login_required
+from django.core.exceptions import PermissionDenied
+from django.core.urlresolvers import reverse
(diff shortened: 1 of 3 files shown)


rdrf/rdrf/reporting_table.py
```diff
@@ -99,20 +111,7 @@ def __init__(self, user, registry_model, multisection_handler, humaniser, type_o
...
-        database = report_db_data["NAME"]
-        host = report_db_data["HOST"]
-        port = report_db_data["PORT"]
-        connection_string = "postgresql://{0}:{1}@{2}:{3}/{4}".format(db_user,
-                                                                      db_pass,
-                                                                      host,
-                                                                      port,
-                                                                      database)
-
-        logger.debug("connection_string = %s" % connection_string)
```
id2359: good idea

rdrf/explorer/views.py
```diff
@@ -190,20 +189,16 @@ def post(self, request, query_id, action):
...
-        spreadsheet_report.run()
-        finish = datetime.now()
-        elapsed_time = finish - start
-        logger.debug("report took %s seconds" % elapsed_time)
-        output = open(spreadsheet_report.output_filename)
-        filename = "Longitudinal Report.xlsx"
-        response = HttpResponse(FileWrapper(output), content_type='application/excel')
-        response['Content-Disposition'] = 'attachment; filename="%s"' % filename
-        return response
+        with NamedTemporaryFile() as output:
```
id2359: Much nicer

PR #312, opened 2016-11-08
aloe test tidy

aahunter:
Minor improvements and fixes for aloe tests.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -474,7 +474,7 @@ dev_aloe() {
 
     # Use run so we can get correct return codes from test run
     set +e
-    _start_aloetests run --rm devaloe
+    _start_aloetests run --rm devaloe $@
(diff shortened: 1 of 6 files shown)


aahunter:
The selenium scrolling to element prior to click comes from stack exchange. It is in multiple posts, I believe this is the one I used in the end:

http://stackoverflow.com/questions/20167544/selenium-scroll-element-into-center-of-view

lib.sh
```diff
@@ -491,7 +491,7 @@ prod_aloe() {
 
     # Use run so we can get correct return codes from test run
     set +e
-    _start_aloetests run --rm prodaloe $@
+    _start_aloetests run --rm prodaloe aloe$@
```
id2359: shouldn't there be a space?

PR #314, opened 2016-11-16
Aloe development fixtures

aahunter:
Simple PR to allow for aloe tests against dev fixtures. Plus don't use runserver_plus for tests - it can swallow some exceptions.


rdrf/rdrf/features/base_admin_page.feature
```diff
@@ -0,0 +1,82 @@
...
+    And I should see a link to "Other Clinicians"
+    And I should see a link to "Doctors"
+    And I should see a link to "Reports"
+    And I should see a link to "Users"
+    And I should see a link to "Genes"
+    And I should see a link to "Laboratories"
+    And I should see a link to "Registries"
+    And I should see a link to "Registry Form"
+    And I should see a link to "Sections"
+    And I should see a link to "Data Elements"
```
id2359: These all look good - they're all fixed - for other menus ( what we call "quick links") - what's shown is dependent on user class / registry features

rdrf/rdrf/features/steps.py
```diff
@@ -214,6 +233,12 @@ def click_save_button(step):
     click(save_button)
 
 
+@step('And I save the form')
+def click_save_button(step):
+    save_button = world.browser.find_element_by_class_name("btn-primary")
```
id2359: More likely to break if we ever introduce another button.

id2359:
ok looks good Adam

rdrf/rdrf/features/steps.py
aahunter: I couldn't see that we are using a consistent way to identify elements. The method directly above, for instance, is doing the exact same thing (clicking save button on form) using a different part of the webdriver api. Its also problematic in that you can't just use the high level natural language feature file when writing the tests - you need to know how the element is identified (class, id, xpath etc) and then what 'natural language' triggers that particular identification. Which defeats the purpose of the natural language. It would be good to talk that through and achieve a consensus on how we want to identify page elements to aid writing tests.

rdrf/rdrf/features/base_admin_page.feature
aahunter: These are a copy of a feature file that was already there. They just run against the dev fixtures rather than dd or fh (can't recall which).

PR #316, opened 2016-11-18
Fhdataimport

id2359:
One off script to import FH patient data from a spreadsheet into RDRF.

This is the first import of non-multisection data - we expect there might be some more "genetic variation" data which is one to many , so I've kept track in the output pairings of there old patient id and ours.

I had two issues:

- Transactions: I was getting an error  so I opted for a pseudo-transaction
- Calculated fields: pushing the spreadsheet value of a calculated field meant that there would be a discrepancy of what was being stored in mongo and what was displayed ( sometimes)  ( a discrepancy that exist until the form was saved) I opted to calculate one of them on the fly (bmi)


rvl:
* rdrf/rdrf/scripts/yaml2excel.py: what are these changes for?

* rdrf/registry/patients/models.py:1177: what kind of exception would be thrown by delete? it should either catch a specific exception or none at all. also, `logger.exception` should be used instead of `logger.error` so that a traceback is logged.
  
* rdrf/rdrf/scripts/data_dictionary_generator.py: not sure what this script is for?
  
* rdrf/rdrf/scripts/data_dictionary_generator.py:141: csv module would have been a better choice. in python3 it even works with unicode data. if there are tab/newline characters in the data they would need to be escaped.
  
* rdrf/rdrf/scripts/data_dictionary_generator.py:269: in python3, you can just use normal `open()` with the `encoding` argument.
  
* rdrf/rdrf/scripts/fh_data_importer.py: not much value in looking at it too carefully because it's a once-off script.

* rdrf/rdrf/scripts/fh_data_importer.py:341: regrettable that transactions didn't work out. you might as well just delete the `use_transactions` flag and conditionals.
  
* rdrf/rdrf/scripts/fh_data_importer.py:741: `evaluate_field_expression()` should be split into two methods -- one to calculate the value, one to update dynamic data.
  
* rdrf/rdrf/scripts/fh_data_importer.py:809: hmm
  
* rdrf/rdrf/scripts/fh_data_importer.py:833: this comment doesn't match the actual command line args.

id2359:
Thanks Rodney

id2359:
I've removed the old yaml2excel script as suggested.
Removed transaction code
Data dictionary script is still used - it allows client to know how to munge their data into the spreadsheet.
I removed the try except on :
            instance.relative_patient.delete()
( This was erroring when transactions were used in the script )

PR #317, opened 2016-11-21
Aloe related bug fixes, tidy and starting to work with multi sections

aahunter:


rdrf/rdrf/features/steps.py
```diff
@@ -107,16 +107,13 @@ def enter_cde_on_form(step, cde_value, form, section, cde):
...
-    utils.click(save_button)
-
-
-@step('And I save the form')
-def click_save_button(step):
-    save_button = world.browser.find_element_by_class_name("btn-primary")
-    utils.click(save_button)
+@step('I click the "(.*)" button')
+def click_submit_button(step, value):
+    """click submit button with given value
```
id2359: this will break if we switch to a different language I think as the value will get translated on the page.

id2359:
approved

PR #319, opened 2016-11-24
models: realign RegistryForm.position

rvl:
Between v0.5.2 and v0.5.3, django-positions changed its field type from PositiveIntegerField to IntegerField, but didn't provide a migration. Result is that the PostgreSQL constraint still exists, even after django-positions is no longer used.

This model change should bring the django field back into line with the table column.

If Django complains about a constraint already existing, then the migration will need to be forced.

Fixes muccg/angelman#162


id2359:
Approved

id2359:
I've merged - cheers

PR #334, opened 2016-11-30
Minor test cleanup

aahunter:
Minor changes primarily to allow for downstream projects to run unique aloe features

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -142,9 +142,6 @@ publish_docker_image)
     publish_docker_image
     ;;
 runtests)
-    create_base_image
-    create_build_image
-    create_dev_image
     run_unit_tests
     ;;
 start_test_stack)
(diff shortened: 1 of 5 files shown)


develop.sh
```diff
@@ -158,18 +158,18 @@ ci_docker_login)
...
-    dev_aloe $@
+    dev_aloe aloe $@
     ;;
 reexport_test_zips)
     reexport_test_zips
     ;;
 prod_aloe)
     shift
-    prod_aloe $@
+    prod_aloe aloe $@
```
id2359: why is aloe needed as an argument

aahunter: Strictly speaking here is isn't needed. But downstream projects can pass different values for this arg. Changing it here so that there is one less point of difference.

PR #360, opened 2017-01-23
Adds possibility to order follow up forms by their name.

sztamas:
The follow up forms are currently ordered by creation time.

This feature allows the registry owners to change the ordering
of the links to the follow up forms by using:

Admin -> Registry Context Form Groups

The ordering can be set to "Creation Time", which is the current
behaviour, or to "Name" which will order the links by their name, which
for FH is the Date of Assessment.


rdrf/rdrf/exporter.py
```diff
@@ -496,6 +496,7 @@ def _get_context_form_groups(self):
             cfg_dict["forms"] = []
             for form in cfg.forms:
                 cfg_dict["forms"].append(form.name)
+            cfg_dict["ordering"] = cfg.ordering
```
id2359: did you make a change to the importer also?

rdrf/rdrf/models.py
```diff
@@ -1949,13 +1960,16 @@ class ContextFormGroup(models.Model):
                       ("N", "Automatic - Number"),
                       ("M", "Manual - Free Text"),
                       ("C", "CDE - Nominate CDE to use")]
+    ORDERING_TYPES = [("C", "Creation Time"),
```
id2359: cool

rdrf/rdrf/models.py
```diff
@@ -2002,6 +2024,21 @@ def get_default_name(self, patient_model):
         else:
             return "Modules"
 
+    def get_value_from_cde(self, patient_model, context_model):
+        form_name, section_code, cde_code = self.naming_cde_to_use.split("/")
+        try:
+            cde_value = patient_model.get_form_value(self.registry.code,
```
id2359: Is it possible to specify a cde in a multisection in the field ?

rdrf/rdrf/exporter.py
sztamas: hm, I was sure that I did and that I tested it, but lost the change somehow.
Thanks for picking it up :)

rdrf/rdrf/models.py
sztamas: well, not sure. I did what the `get_name_from_cde` below was doing.
Should these be changed?

id2359: yeah I don't think I added any validation at all for that - so leave it as is

id2359:
approved from me

Issue #287, opened 2016-10-24
Styling of 'delete' button 


knapier:
Using chrome, delete button appears cut off

<img width="1222" alt="screen shot 2016-10-24 at 10 06 17 am" src="https://cloud.githubusercontent.com/assets/17611855/19631848/4f4b4854-99d3-11e6-9792-6c34ef5800a2.png">

rvl:
Does this look like an ok fix?

![admin-buttons](https://cloud.githubusercontent.com/assets/1019641/19677633/21d352c6-9a92-11e6-9b26-0bd202b2fe97.png)

knapier:
Looks good to me @rvl !

PR #361, opened 2017-01-25
Patient deletion

id2359:
Summary of work done:

1) Made patient deletion always soft delete - added a _hard_delete method on patient that is only called from a management command
2) Made normal patient model manager only retrieve non-archived patients - archived patients do not appear in the forms GUI or patient listing  --edit : They are listable by admin from the Admin Page as "Archived Patients" - from there, hard delete , unarchiving is possible 
3) Added a management command (rdrf_archiver)  to list archived patients and hard delete , unarchive ,archive patients
4) Added  a modal delete button to the banner for all patient views which deletes via REST api we have exposed, after deletion of a patient the user is redirected back to the patient listing.

FH specific ( actually specific to patients with linkages to other patients via family linkage) 

5) Added a check to this modal for FH where family linkage allows patients to be related to each other: Archiving of index patients is not possible - we indicate to the user than a new index for a given family must first be "promoted" in the family linkage screen
6) Added logic to sever the link between "Relative objects" and the patient created from them when archived 

Known issues
I've yet to add a an aloe test - will have a look also today

The modeladmin for ArchivedPatient shows an "Add" button in the top right of the admin form -- I have removed add_permission on this model admin but annoyingly the button is still clickable/visible

I've made a mistake in my commit messages 
should have been rdrf-ccg #287 not rdrf #287
https://github.com/muccg/rdrf-ccg/issues/287


rdrf/rdrf/templates/rdrf_cdes/form.html
```diff
@@ -236,7 +236,7 @@
 {% endblock %}
 
 {% block formbtns %}
-	<br>
+        <!-- xxx -->
```
sztamas: is this still needed or was used to debug?

id2359: was used to debug I'll remove it

PR #362, opened 2017-01-25
Aloe prod named volume

aahunter:
Updated docker-compose yml files to version 2, use a named volume for local prod uwsgi that corrects a permissions problem that was breaking ci builds .


id2359:
approved

Dockerfile-builder
```diff
@@ -24,6 +24,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 COPY docker-entrypoint-build.sh /app/docker-entrypoint-build.sh
 
+RUN mkdir -p /data \
+    && /bin/chown ccg-user:ccg-user /data
```
id2359: ok

Issue #369, opened 2017-01-31
The users should be warned that their password will expire soon


sztamas:
Currently, there is no warning for the user that their password will expire, so if they are not proactive about changing their password regularly one day they won't be able to log in.

The details is to be determined, can be email and/or warning on the web page.

sztamas:
For now I've just added a warning message to the web page.
I think that's good enough and doesn't need the complications of email templates having to be configured etc.

Let me know if the email notification is a must as well.

knapier:
Email notification would be good, and has been requested by FH. I'll ticket up.

Issue #368, opened 2017-01-31
Make it obvious why the user has been deactivated 


sztamas:
Currently, if the user has been deactivated by django-useraudit there is no easy way to know by the admin or the user why that happened (ie. did the password expire?, has the user not logged in and the account has been deactivated? etc.)

We should change the admin user view to show the reason and possibly the login page as well to show the reason to the user, because for password expiry the user could solve the problem by themselves just by resetting the password.

Some of these changes will probably need changes in django-useraudit.

sztamas:
Changed the admin page of users list to include the reason why the user has been deactivated if the reason is known.

PR #370, opened 2017-01-31
User deactivation pass expiry improvements

sztamas:
Addresses #368 and #369.

### DIFF
--- a/rdrf/registry/groups/admin.py
+++ b/rdrf/registry/groups/admin.py
@@ -7,7 +7,7 @@ from django.core.exceptions import ValidationError
 
 from .admin_forms import UserChangeForm, RDRFUserCreationForm
 from .models import WorkingGroup
-from useraudit.password_expiry import is_password_expired
+from useraudit.models import UserDeactivation
 
 import logging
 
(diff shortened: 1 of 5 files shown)


rdrf/registry/groups/admin.py
```diff
@@ -121,16 +121,18 @@ def get_registries(self, obj):
...
-                        args=(obj.pk,)))
-        return "Yes" if obj.is_active else "No"
-    is_active_expired.short_description = "Is active"
+    def status(self, user):
+        if user.is_active:
+            return 'Active'
+        choices = dict(UserDeactivation.DEACTIVATION_REASON_CHOICES)
+        last_deactivation = UserDeactivation.objects.filter(username=user.username).order_by('-timestamp').first()
+        if last_deactivation is None or last_deactivation.reason not in choices:
+            return 'Inactive'
```
id2359: Why would last_deactivation.reason not be in choices?

id2359:
Approved by me

rdrf/registry/groups/admin.py
sztamas: Wrong value in the DB, out of sync with choices. Shouldn't happen "normally".

PR #373, opened 2017-02-01
Aloe: bail out early if importing the zip files fails

### LINKED ISSUE #371: If test zips are bad in some way - bail out of aloe tests early
E.g. When schema differences are detected in the zips and running RDRF - the import fails but the tests proceed ( and fail slowly)

Fail quickly instead

sztamas:
Fixes #371

### DIFF
--- a/rdrf/rdrf/features/utils.py
+++ b/rdrf/rdrf/features/utils.py
@@ -130,7 +130,7 @@ def load_export(export_name):
 
 
 def django_import(export_name):
-    django_admin(["import", "{0}/{1}".           format(exported_data_path(), export_name)])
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/features/utils.py
```diff
@@ -149,9 +149,12 @@ def django_migrate(args=[]):
     django_admin(["migrate", "--noinput"] + args)
 
 
-def django_admin(args):
+def django_admin(args, fail_on_error=False):
     logger.info(args)
-    subprocess_logging(["django-admin.py"] + args)
+    return_code, _, _ = subprocess_logging(["django-admin.py"] + args)
```
id2359: Looks good

PR #374, opened 2017-02-02
Shell linting

aahunter:
Linting docker-entrypoint.sh

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -270,7 +270,7 @@ if [ "$1" = 'aloe' ]; then
     _aloe "$@"
 fi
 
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -137,21 +177,24 @@ function _django_fixtures {
 
 
 function _runserver() {
-    echo "RUNSERVER_OPTS is ${RUNSERVER_OPTS}"
+    : "${RUNSERVER_OPTS=${RUNSERVER_CMD} 0.0.0.0:${RUNSERVERPORT} --settings=${DJANGO_SETTINGS_MODULE}}"
```
id2359: So this is more than linting as such, right?

aahunter: Removing duplicate declarations of the same var can probably count as linting - this snippet doesn't really show that though. FWIW it wasn't identified by the linter as such.

PR #379, opened 2017-02-06
aloe: reset password change data on exported data import

aahunter:
Trivial change for test exported data - reset password change date.

### DIFF
--- a/rdrf/rdrf/features/utils.py
+++ b/rdrf/rdrf/features/utils.py
@@ -22,6 +22,13 @@ def stellar_config_path():
     return os.path.join(utils_path(), 'stellar')
 
 
+def reset_password_change_date():
+    logger.info('')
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/features/utils.py
```diff
@@ -22,6 +22,13 @@ def stellar_config_path():
     return os.path.join(utils_path(), 'stellar')
 
 
+def reset_password_change_date():
+    logger.info('')
+    from registry.groups.models import CustomUser
+    from django.utils import timezone
+    CustomUser.objects.update(password_change_date=timezone.now())
```
id2359: looks good

id2359:
empty log message

id2359:
approved although we'd need to test with the original "bad" test zips to show it fixes the issue

Issue #393, opened 2017-02-13
Modjgo management command converted dates to isostrings 


id2359:

id2359:
Modjgo objects created foe the following mongo collections will contain date cdes : cdes, history

PR #395, opened 2017-02-15
Datecdesdatamigration

id2359:
This is no longer a data migration despite the branch name.
It is a script which should be run on the live site ( post a 2.0 install)
The script runs over Modjgo objects ( either cdes or history) and updates the data if it detects any "bad" date cde values. Only values of common data elements of type date are updated.
The script creates an instance of a TimesStripper class or HistoryTimeStripper class defined in utils.
There are unit tests on this class - initially I was thinking this would work in a data migration
so I added a backward step, but now I am only using the forward step in a transaction in the script.


id2359:
ticket https://github.com/muccg/rdrf/issues/393

rdrf/rdrf/tests.py
```diff
@@ -485,3 +486,569 @@ def test_date(self):
...
+            def __init__(self, pk, data):
+                self.pk = pk
+                self.data = data
+            def save(self):
+                print("Fake Modjgo save called")
+
+
+class TimeStripperTestCase(TestCase):
+    def setUp(self):
+        super(TimeStripperTestCase, self).setUp()
```
sztamas: The super(CLASS, self).x(...) calls can all be changed to super().x(...) now that we're on python 3. 

Applies to all the similar calls.

rdrf/rdrf/utils.py
```diff
@@ -441,3 +442,164 @@ def error(text):
...
+                            items = section["cdes"]
+                            for item in items:
+                                for cde in item:
+                                    if self.is_date_cde(cde):
+                                        if self.update_cde(cde):
+                                            updated += 1
+                                            
+        return updated > 0
+                    
+    def backward(self):
```
sztamas: this can be deleted as discussed on Slack.

rdrf/rdrf/utils.py
```diff
@@ -441,3 +442,164 @@ def error(text):
...
+    moving from Mongo to pure Django models with JSON fields ( "Modjgo" 0objects.)
+    CDE date values were converted  into iso strings including a time T substring.
+    This was done recursively for the cdes and history collections 
+    We provide forwards and backward methods to be used in a migration to fix this
+    I have used a class so we can unit test more easily.
+    """
+    
+    def __init__(self, dataset):
+        self.dataset = dataset # queryset live , lists of data records for testing
+        self.backup_data = {}
```
sztamas: I think the contents of this (what has been updated) could be written out to a file after the script finished, so we have a record of what exactly has been updated.

A bit of extra precaution doesn't hurt.

rdrf/rdrf/utils.py
```diff
@@ -441,3 +442,164 @@ def error(text):
 
     def __repr__(self):
         return self.text
+
+
+
+class TimeStripper(object):
+    """
+    This class exists to fix an error we introduced in the migration
+    moving from Mongo to pure Django models with JSON fields ( "Modjgo" 0objects.)
```
sztamas: Also, can you please update the comment to not say this should be used from a migration.

Issue #412, opened 2017-02-20
Demographics- Formatting 


id2359:
see https://github.com/muccg/rdrf-ccg/issues/871

id2359:
fixed in lee branch will create pull request

Issue #398, opened 2017-02-16
Patient List- sorting by working groups does not work


id2359:

id2359:
https://github.com/muccg/rdrf-ccg/issues/873

id2359:
fixed in lee branch

PR #413, opened 2017-02-20
patient list sorting and demographics formatting

id2359:
Fixed for tickets #398 and #412


rdrf/rdrf/form_view.py
```diff
@@ -2130,8 +2130,9 @@ def post(self, request, registry_code, patient_id, context_id=None):
             logger.debug("******************** saving custom consent form *********************")
             custom_consent_form.save()
             logger.debug("******************** end of consent save *********************")
-            context["message"] = "Consent details saved successfully"
-            return HttpResponseRedirect(self._get_success_url(registry_model, patient_model))
+            context["message"] = "Patient %s %s saved successfully" % (patient_model.given_names,
+                                                                       patient_model.family_name)
+            #return HttpResponseRedirect(self._get_success_url(registry_model, patient_model))
```
aahunter: minor point - remove the commented out cruft

id2359:
I'll write some aloe tests as discussed in the meeting

rdrf/rdrf/patients_listing.py
```diff
@@ -455,6 +475,7 @@ def icon(self, tick):
 
 class ColumnWorkingGroups(Column):
     field = "working_groups_display"
+    sort_fields = ["working_groups_display"]
```
sztamas: I might be missing something but can't we just use `working_groups__name` here and drop all the special case translating code above?

rdrf/rdrf/patients_listing.py
```diff
@@ -264,7 +264,17 @@ def apply_custom_ordering(self, qs):
         if key_func:
             # we have to retrieve all rows - otherwise , queryset has already been
             # ordered on base model
-            return sorted(qs, key=key_func[0], reverse=(self.sort_direction == "desc"))
+            logger.debug("key = %s" % key_func[0])
+
+            def key_func_wrapper(thing):
+                value = key_func[0](thing)
+                if value is None:
+                    return ""
```
sztamas: Are all the column values strings?

Won't this fail for ints, dates, etc. because we're comparing incompatible types?

id2359: That is correct so if None was in a column for a date it would fail - we're mandating date of birth though. 
I could make a type map of the columns I suppose

id2359: diagnosis updated < 365  is a string ( it works) 

but currency of genetic data is bool so fails

so I'll add a  map

id2359: There's a better way involving  a "bottom" type or MinType lass which I've used

aahunter:
:+1:  LGTM

PR #419, opened 2017-02-28
rdrf-ccg #899 Modified cde history popup functions to use Modjgo

id2359:
In short : Modjgo objects being returned on some queries broke the history popup.

### DIFF
--- a/rdrf/rdrf/dynamic_data.py
+++ b/rdrf/rdrf/dynamic_data.py
@@ -474,21 +474,33 @@ class DynamicDataWrapper(object):
         else:
             return nested_data
 
-    def get_cde_val(self, registry_code, form_name, section_code, cde_code):
-        data = self._get_record(registry_code, "cdes")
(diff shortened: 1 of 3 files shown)


rdrf/rdrf/dynamic_data.py
```diff
@@ -474,31 +474,47 @@ def load_dynamic_data(self, registry, collection_name, flattened=True):
...
                 return diff
             return list(filter(is_different, snapshots))
+        
         record_query = self._get_record(registry_code, "history", filter_by_context=False)
         record_query = record_query.find(record_type="snapshot")
-        data = map(fmt, record_query.data())
+        data = []
+        for i, snapshot in enumerate(record_query.data()):
+            data.append(fmt(snapshot, i))
+
```
sztamas: Why not list comprehension? Would be cleaner and shorter.

id2359: yeah true ok

PR #424, opened 2017-03-09
File link display+ tests 

id2359:
File CDEs allow a file to uploaded.
When uploaded a download  link should appear at the file cde location.
An uploaded file can be replaced by uploading another file.
An uploaded file can be deleted by checking a "clear" checkbox near the file link - when checked , the file storage module deletes the file.

Changes have been made to the underlying file storage mechanism which resulted in several issues 

The file link for a download was not displaying in the view's post response. ( this was a logic error in the view not the file storage change: the "wrapping" of file widgets was occurring too early so that no upload ids had been created at that stage) The fix was to do recreate the wrapped section forms at a later stage when the upload ids had been generated.

The clear checkbox didn't work and caused a runtime error ( filestorage api signature change)
Reuploading was broken ( ditto)

some aloe tests were written to test:

Uploading a file and observing that a download link was visible and of the right form ( /uploads/<upload_id> where the upload id is a django model pk - I test for the pattern because this indirectly checks that a model object has been created server side 

Replacing a file : same idea

I tried to get a delete/clear test working but couldn't so I've commented it out for now.

### DIFF
--- a/rdrf/rdrf/dynamic_data.py
+++ b/rdrf/rdrf/dynamic_data.py
@@ -598,7 +598,7 @@ class DynamicDataWrapper(object):
         elif is_uploaded_file(value):
             # A file was uploaded.
             # Store file and convert value into a file wrapper
-            to_delete = current_value
+            to_delete = False
             value = filestorage.store_file_by_key(registry_code, None, key, value)
         else:
             to_delete = None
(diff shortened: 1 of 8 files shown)


rdrf/rdrf/dynamic_data.py
```diff
@@ -586,7 +586,8 @@ def handle_file_upload(registry_code, key, value, current_value):
             # Django uses a "clear" checkbox value of False to indicate file should be removed
             # we need to delete the file but not here
             logger.debug("User cleared %s - file will be deleted" % key)
-            to_delete = current_value
+            to_delete = True
```
aahunter: Can we make some minor improvements to the legacy code here:

1. Change to_delete as always Boolean. (Sometimes set as None)
2. Don't redefine value, use a separate rval var defined at the top of the function
3. Define to_delete as False at he top of the function, function scope, then we can remove the final else

These aren't comments on your code, rather on the legacy code, but it will only take a few mintues and we might as well tighten that up.

cheers.

PR #427, opened 2017-03-14
Mark for deletion

id2359:
Items ( blocks of CDEs) in a multisection can be "marked for deletion" and then the form saved, causing the entire block of cdes ( form in a formset)  to be deleted - this has the effect of changing the indices of the items. In order to correctly display file links for file cdes on cdes which have not been uploaded in the POST ( but have in earlier POSTS) - values of the uploaded file django ids for the cde need to be retrieved but IF items have been marked for deletion ( and deleted in the current POST) the cde values from the correct index needs to be loaded prior to the record being saved

An "index map" of new indices to old indices after a deletion was introduced to enable this , but it wasn't being correctly threaded some of the refactored code for the file storage code resulting in the wrong item's data being presented in the view.

I've added an aloe test to upload files in 2 multisection items and then perform a delete of the first item.

The test checks that the second upload file is available but the first isn't.

### DIFF
--- a/rdrf/rdrf/features/fh_multisection_tests.feature
+++ b/rdrf/rdrf/features/fh_multisection_tests.feature
@@ -32,8 +32,8 @@ Feature: User operates on multisection items.
     #And I enter value "3.0" for form "Imaging" section "Carotid Ultrasonography" cde "Result (left)"
     And I enter value "3.0" for form "Imaging" section "Carotid Ultrasonography" cde "Result (right)" in item 2
     And I enter value "item 2" for form "Imaging" section "Carotid Ultrasonography" cde "Result" in item 2
-    When I upload file "/app/rdrf/rdrf/features/fh_multisection_tests.feature" for multisection "Carotid Ultrasonography" cde "Report" in item 2
+    When I upload2 file "/app/rdrf/rdrf/features/fh_multisection_tests.feature" for multisection "Carotid Ultrasonography" cde "Report" in item 2
     And I click the "Save" button
     Then I should see "Patient John SMITH saved successfully"
     And I should be able to download "fh_multisection_tests.feature"
-    #And I should be able to download "fh_file_upload.feature"
+    And I should be able to download "fh_file_upload.feature"
--- a/rdrf/rdrf/features/fh_multisection_tests.feature
+++ b/rdrf/rdrf/features/fh_multisection_tests.feature
@@ -1,0 +1,39 @@
+Feature: User operates on multisection items.
+  A user can:
+  Add items (single section data) to a multisection and delete an item
+  by checking the "Mark for Deletion" checkbox and saving.
+  
+  Background:
+    Given export "fh.zip"
+    Given a registry named "FH Registry"
+  
+  Scenario: Add multiple items to a multisection.
+    When I am logged in as curator
+    When I click Module "Main/Imaging" for patient "SMITH John" on patientlisting
+    Then location is "Main/Imaging"
+  
+    # Enter first item
+    #When I click radio button value "Yes - Normal" for section "Carotid Ultrasonography" cde "Carotid Ultrasonography"
+    And I enter value "01-01-2017" for form "Imaging" section "Carotid Ultrasonography" cde "Date"
+    # for some reason the line below failed
+    #And I enter value "3.0" for form "Imaging" section "Carotid Ultrasonography" cde "Result (left)"
+    And I enter value "4.0" for form "Imaging" section "Carotid Ultrasonography" cde "Result (right)"
+    And I enter value "item 1" for form "Imaging" section "Carotid Ultrasonography" cde "Result"
+    When I upload file "/app/rdrf/rdrf/features/fh_file_upload.feature" for multisection "Carotid Ultrasonography" cde "Report"
+    And I click the "Save" button
+    Then I should see "Patient John SMITH saved successfully"
+    And I should be able to download "fh_file_upload.feature"
+
+    # Enter second item
+    #When I click radio button value "Yes - Normal" for section "Carotid Ultrasonography" cde "Carotid Ultrasonography"
+    When I click the add button for multisection "Carotid Ultrasonography"
+    And I enter value "02-01-2017" for form "Imaging" section "Carotid Ultrasonography" cde "Date" in item 2
+    # for some reason the line below failed
+    #And I enter value "3.0" for form "Imaging" section "Carotid Ultrasonography" cde "Result (left)"
+    And I enter value "3.0" for form "Imaging" section "Carotid Ultrasonography" cde "Result (right)" in item 2
+    And I enter value "item 2" for form "Imaging" section "Carotid Ultrasonography" cde "Result" in item 2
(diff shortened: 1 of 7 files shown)


rdrf/rdrf/features/steps.py
```diff
@@ -468,6 +565,11 @@ def upload_file(step, upload_filename, section, cde):
     input_element = scroll_to_element(step, section, cde)
     input_element.send_keys(upload_filename)
 
+@step('upload2 file "(.*)" for multisection "(.*)" cde "(.*)" in item (\d+)')
```
sztamas: Can't that be just upload instead of upload2?

rdrf/rdrf/form_view.py
```diff
@@ -502,9 +524,22 @@ def post(self, request, registry_code, form_id, patient_id, context_id=None):
 
                     sections_to_save.append(section_info)
 
-                    current_data = dyn_patient.load_dynamic_data(self.registry.code, "cdes")
 
-                    form_data = wrap_file_cdes(registry_code, dynamic_data, current_data, multisection=True)
+                    form_data = wrap_file_cdes(registry_code, dynamic_data, current_data, multisection=True, index_map=index_map)
+                    logger.debug("form_data on the form for section %s  = %s" % (s, form_data)) 
+                    try:
```
sztamas: In general, not sure if you want to leave all that detailed logging that has been added, but especially true for this block of code. 
Does it have to stay?

rdrf/rdrf/features/steps.py
```diff
@@ -581,3 +699,61 @@ def clear_file_upload(step, section, cde, download_name):
...
+
+@step('I click the add button for multisection "(.*)"')
+def add_multisection_item(step, section):
+  xpath = ".//div[@class='panel-heading' and contains(.,'%s') and not(contains(., '__prefix__')) and not(contains(.,'View previous values'))]" % section
+  div = world.browser.find_element_by_xpath(xpath)
+  add_link_xpath = """.//a[starts-with(@onclick,"add_form('formset_")]"""
+  add_link = div.find_element_by_xpath(add_link_xpath)
+  add_link.click()
+  # sometimes the next cde send keys was going to the wrong item
+  wait_n_seconds(step, 5)
```
sztamas: Not sure why this explicit sleep is needed, but in general I always ask if they can be replaced with waiting for something to appear etc. with a timeout.

rdrf/rdrf/form_view.py
id2359: yeah I guess not - I'll remove it

rdrf/rdrf/file_upload.py
```diff
@@ -59,6 +59,10 @@ def __str__(self):
         """
         return self.fs_dict['file_name']
 
+    def __unicode__(self):
```
id2359: yeah I added it mistakingly

rdrf/rdrf/features/steps.py
```diff
@@ -41,9 +41,58 @@ def scroll_to(element):
      scroll_to_y(y)
      print("scrolled to %s at (0,Y) = %s" % (moniker(element), y))
      return y
+
+
+def scroll_to_multisection_cde(section, cde, item=1):
+    # item 1 means the 1st block of cdes in the multisection
+    print("Attempting to scroll to section %s cde %s item %s" % (section,
```
aahunter: Can we use the logger?

rdrf/rdrf/features/steps.py
```diff
@@ -41,9 +41,58 @@ def scroll_to(element):
...
+    
+    for label_element in default_panel.find_elements_by_xpath(label_expression):
+        print("found a label element for cde %s" % cde)
+        input_div = label_element.find_element_by_xpath(".//following-sibling::div")
+        try:
+            input_element = input_div.find_element_by_xpath(".//input[contains(@id, '%s')]" % formset_string)
+        except:
+            input_element = None
+            
+        if not input_element:
```
aahunter: Minor point but I would flip this conditional. ie the norm is to continue the loop and the exception is to return.

rdrf/rdrf/features/steps.py
```diff
@@ -183,6 +243,43 @@ def enter_cde_on_form(step, cde_value, form, section, cde):
...
+        ".//div[@class='panel-heading'][contains(., '%s')]" % section)
+    section_div = section_div_heading.find_element_by_xpath("..")
+
+    label_expression = ".//label[contains(., '%s')]" % cde
+
+    for label_element in section_div.find_elements_by_xpath(label_expression):
+        input_div = label_element.find_element_by_xpath(".//following-sibling::div")
+        try:
+            input_element = input_div.find_element_by_xpath(".//input")
+            if not correct_item(input_element):
```
aahunter: same as above, I would invert he logic here to remove the continue and return based on an if

rdrf/rdrf/features/steps.py
```diff
@@ -41,9 +41,58 @@ def scroll_to(element):
      scroll_to_y(y)
      print("scrolled to %s at (0,Y) = %s" % (moniker(element), y))
      return y
+
+
+def scroll_to_multisection_cde(section, cde, item=1):
```
aahunter: Can we rename scroll_to_multisection_cde to find_multi_section_cde (or similar) and remove the scroll_to call. I think we should separate those two functions - finding things and clicking on them. Both can be a bit tedious to debug in aloe/selenium.

Related: In case it was missed: there is a utils.click method that scrolls to items before clicking.

I realise there are not in the PR (earlier commits) but if the scroll_to() methods are needed can we move them to utils.py, it might also be a good idea to add a note about why they are needed / or under what conditions the other scroll/click methods don't work.

id2359: will do

rdrf/rdrf/features/fh_multisection_tests.feature
```diff
@@ -0,0 +1,43 @@
...
+  Scenario: Add multiple items to a multisection.
+    When I am logged in as curator
+    When I click Module "Main/Imaging" for patient "SMITH John" on patientlisting
+    Then location is "Main/Imaging"
+  
+    # Enter first item
+    And I enter value "01-01-2017" for form "Imaging" multisection "Carotid Ultrasonography" cde "Date" in item 1  
+    And I enter value "4.0" for form "Imaging" multisection "Carotid Ultrasonography" cde "Result (right)" in item 1
+    And I enter value "item 1" for form "Imaging" multisection "Carotid Ultrasonography" cde "Result" in item 1
+    When I upload2 file "/app/rdrf/rdrf/features/fh_file_upload.feature" for multisection "Carotid Ultrasonography" cde "Report" in item 1
```
aahunter: Apologies for the duplicate comment - I accidentally put this comment on the commit also.

Can we remove the path? In utils there is a "def utils_path():", that allows you to get the path to utils at runtime, perhaps pick a file relative to that location. The paths for stellar are constructed this way, for example.

It avoids having a hard coded path that will only work from within a docker container (admittedly, this is what we are doing 99.99% of the time)

PR #436, opened 2017-04-27
I18n

aahunter:
This is internationalisation work by @id2359 .


aahunter:
I think this is an issue from previous work (Don't see it in the changelog, but the result is a single char)

#: rdrf/rdrf/components.py:169
msgid "s"
msgstr ""

aahunter:
Another observation (looking at a yaml file in a downstream project):

```
- header: |-
    <p><strong><u>Respiratory function</u>
    <br>
    Ventilation</strong> means breathing support from a mechanical ventilation device in the form of either non-invasive ventilation via a face or nose mask, or invasive ventilation via a tracheostomy (an operation to make an incision in the wind-pipe) or endotracheal tube (a breathing tube inserted into the wind-pipe). Ventilatory support can be used either all day or just for a few hours.</p>
```

Yields (partially):

```
msgid ""
"Ventilation"
msgstr "translation goes here"
msgid " means breathing support from a mechanical ventilation device in the form of either non-invasive ventilation via a face or nose mask, or invasive ventilation via a tracheostomy (an operation to make an incision in the wind-pipe) or endotracheal tube (a breathing tube inserted into the wind-pipe). Ventilatory support can be used either all day or just for a few hours."
msgstr "TRANSLATION 103"
```

I think this might be problematic - that Ventilation string will be de-duped for example, which may or may not make sense depending of the grammatical structure of the translated language. We may need to make the complete header a translation string.

This is one of several examples of this type of behaviour.

aahunter:
I suspect email templates may be an issue - again probably need a single translation per email template with language set per user settings.

rdrf/rdrf/utils.py
```diff
@@ -591,57 +591,106 @@ def __le__(self, other):
...
+    # cde field:
+    # <reg code>/<form name>/<section code>/<cde code>/<cde field>
+    # section field:
+    # <reg code>/<form name>/<section code>/<section field>
+    # form field:
+    # <reg code>/<form name>/<form field>
+    # registry field
+    # <reg code>/<registry field>
+    # for consents , I have  introduced a more general notion:
+    # model/<ModelName>/<pk>/<fieldname> - delegated to get_field_from_model above
```
aahunter: I like the use of a sentinel at the beginning of the field_path to denote the type of field path as is the case with 'model'. I think an improvement would be to use that everywhere, consistently, not just with model as a special case.

For example:
 # cdefield/<reg code>/<form name>/<section code>/<cde code>/<cde field>
 # sectionfield/<reg code>/<form name>/<section code>/<section field>
 # formfield/<reg code>/<form name>/<form field>
 # registryfield/<reg code>/<registry field>

That would then enable the conditional logic to tighten as well (not needing to count args etc)

rdrf/rdrf/template_loaders/translation.py
```diff
@@ -0,0 +1,46 @@
+import logging
```
aahunter: What about changing the name to something unrelated to i18? I see this as a rdrf specific template loader - nothing more. Along that same thread, I don't see why PRELUDE is needed. That doesn't occur to me to be a concern of the template loader.

rdrf/rdrf/template_loaders/translation.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    A custom template loader which retrieves embedded html templates from the registry definition
+    models.
+    """
+    PREFIX = "rdrf://"
+    PRELUDE = "{% load i18n %}"
+    
+    def get_contents(self, origin):
+        name = origin.name
+        if not name.startswith(self.PREFIX):
```
aahunter: Just a thought - if the template'path' is an URI, then why not parse as a URI (to validate) and check the scheme is 'rdrf'. self.PREFIX becomes self.SCHEME?

rdrf/rdrf/templatetags/get_info_text_expression.py
```diff
@@ -9,22 +8,25 @@
 
 
 @register.filter()
-def get_information_text(fields):
+def get_info_text_expression(fields):
```
aahunter: Minor point coming to this cold I found the method name a little misleading. It gets the path (URI) to a template rather than getting info text.

Also, a bit worried that this abstracts too much. It might be better if the URI was in the actual template and not hidden behind the filter. So the template code can be something like:

`{% include "rdrf://model/ConsentSection/%s/information_text" with consent_section=get_consent_section %}
`
(Forgive me if I munged the django template syntax)

rdrf/rdrf/management/commands/create_translation_file.py
```diff
@@ -0,0 +1,305 @@
...
+                self._extract_html_strings(file_name)
+
+    def _extract_html_strings(self, yaml_file):
+        # This dumps the embedded html templates from the yaml into a temporary folder and
+        # and then runs makemessages over it to extract the strings into the 
+        # into the "system" po file
+        self._load_yaml_file(yaml_file)
+        htmls = []
+
+        # splash screen
```
sztamas: these comments don't add much value IMO

rdrf/rdrf/management/commands/create_translation_file.py
```diff
@@ -0,0 +1,305 @@
...
+    def _load_yaml_file(self, file_name):
+        with open(file_name) as f:
+            try:
+                self.data = yaml.load(f)
+            except Exception as ex:
+                print("could not load yaml file %s: %s" % (file_name,
+                                                           ex))
+
+                sys.exit(1)
+
```
sztamas: just raise `CommandError` with error message.

rdrf/rdrf/management/commands/create_translation_file.py
```diff
@@ -0,0 +1,305 @@
...
+import sys
+import os.path
+import re
+from rdrf.utils import de_camelcase
+from rdrf.models import Registry
+from tempfile import TemporaryDirectory
+from django.core.management import BaseCommand
+from django.core.management import call_command
+
+class Command(BaseCommand):
```
sztamas: For the class:
  * it isn't very clear to me how to use the command. Maybe a bit more of high level documentation or usage notes?
  * IMO, too much is handled in this class

rdrf/rdrf/settings.py
```diff
@@ -53,11 +53,17 @@
 LANGUAGE_CODE = env.get("language_code", 'en')
 USE_I18N = env.get("use_i18n", True)
 
-LANGUAGES = (
-    ('ar', 'Arabic'),
-    ('de', 'German'),
-    ('en', 'English'),
-)
+# This must be a superset of LANGUAGES
```
sztamas: Why is this needed? More doc?
It says that it has to be a superset of `LANGUAGES`, but `LANGUAGES` is read from the environment and the "superset relationship" isn't verified.
Is that important?

rdrf/rdrf/utils.py
```diff
@@ -570,25 +579,60 @@ def munge_data(self, data):
...
+
+def get_registry_definition_value(field_path):
+    # find a value in a registry definition
+    # model/<ModelName>/<pk>/<fieldname> - delegated to get_field_from_model above
+
+    from rdrf.models import Registry
+    from rdrf.models import RegistryForm
+    from rdrf.models import Section
+    from rdrf.models import CommonDataElement
+
```
sztamas: Unused imports

rdrf/rdrf/utils.py
```diff
@@ -570,25 +579,60 @@ def munge_data(self, data):
...
+        parts = model_path.split("/")
+        model_name = parts[1]
+        pk = int(parts[2])
+        field = parts[3]
+        model_class = apps.get_model('rdrf', model_name)
+        model_instance = model_class.objects.get(pk=pk)
+        value = getattr(model_instance, field)
+        return value
+    except Exception as ex:
+        logger.error(
```
sztamas: use `logger.exception` instead?

rdrf/rdrf/utils.py
id2359: ta

rdrf/rdrf/utils.py
id2359: ta

rdrf/rdrf/settings.py
id2359: Originally we had one languages list - the EmailTemplate model field languages using it to filla  dropdown when we exposed to settings / env we created a situation where different deployments would trigger migration errors.
\
To avoid error I created a superset


yes I guess we could  check

id2359:
I've changed management command to address some of Tamas' points so will now merge

PR #446, opened 2017-05-26
Various enhancements to RDRF's handling of internationalisation

id2359:
This branch includes enhancements to enable:
Javascript (client side ) translations via gettext calls.
Links to the correct information pdfs based on the language request header.
Preferred language field on user which is exposed to patient/parent  registration in mtm and in the RDRF admin .
Fix to the OtherPleaseSpecify widget which wasn't correctly translating.
Email notifications workflow corrected to pick the right language template for each recipient based on their preferred language ( not taking it from the request as before.).

Changes utilising the above also exist in mtm codebase


rdrf/rdrf/email_notification.py
```diff
@@ -17,55 +16,111 @@ class RdrfEmail(object):
...
+        if email_notification:
+            self.email_notification = email_notification
+            self.reg_code = self.email_notification.registry.code
+            self.description = self.email_notification.description
+            
+        else:
+            self.email_notification = self._get_email_notification()
+            
 
     def send(self):
```
sztamas: Maybe it is just me, but this `RdrfEmail` abstraction got a bit messy for me.
It looked to me like it represented just one instance of an email we send out, but now from the `send` we can send out multiple emails. There is also some skipping at line 51, and comments says resend will happen per language template which isn't clear how it happens.

Can't we just at the notification level: 
  * prepare all the emails that have to be sent out
  * iterate through that list and create one `RdrfEmail` for each email and send it out.

That will separate concerns a bit better and will keep the `RdrfEmail.send()` simple as it was.

rdrf/rdrf/backends.py
```diff
@@ -19,8 +19,7 @@ def account_lockout_handler(sender, user=None, **kwargs):
     registry_code = getattr(settings, "FALLBACK_REGISTRY_CODE", None)
 
     if registry_code:
-        process_notification(registry_code,"account-locked",
-                             get_request().LANGUAGE_CODE, template_data)
+        process_notification(registry_code,"account-locked", template_data)
```
sztamas: use `EventType.ACCOUNT_LOCKED` here?

id2359: yup I missed it

rdrf/rdrf/email_notification.py
id2359: yeah for "Accept-Language: en-US,en;q=0.5" string I think the code will split on the hyphen and take the first member of the resulting list  ( from memory - I will test it )

id2359: The code before this change had one recipients list built up from the recipient field and the group recipient field and one send ( to those recipients) - with the language being determined from the request )  - I'll take a look again --  
there is the resend functionalitty via the email notification history to take into account as well - ( I did actually refactor this  to make it work) - the original resend was at the level of the one email notifcation ( ie all recipients would receive the communication (again))  - currently it's at the language level( I didn't have to change the ENH listing view at all for this) if we make it at the level of user , we might have to change the gui further

id2359:
Response to review:

- I've removed the custom parsing of the Accept Language header i favour of using Django's util function 

- Consistently used EventTypes

I haven't modified the RdrfEmail model or notification history - I think refactoring this should be triaged in a sep ticket because as it stands the resend work flow isn't actively used according to Kathryn (The issue being resending the account activation is useless if the token has expired)

PR #456, opened 2017-06-22
Qcreatepatienterror

id2359:
Removed chunks of code that are no longer used and caused an error by calling into removed function as part of the questionnaire "rollback" on failure.

### DIFF
--- a/rdrf/rdrf/questionnaires.py
+++ b/rdrf/rdrf/questionnaires.py
@@ -299,7 +299,7 @@ class QuestionnaireReverseMapper(object):
                     k)
                 original_form_name, original_section_code = self.parse_generated_section_code(
                     generated_section_code)
-                reg_code = self.registry_code
+                reg_code = self.registry.code
                 q_data = self.questionnaire_data[k]
                 yield reg_code, original_form_name, original_section_code, cde_code, q_data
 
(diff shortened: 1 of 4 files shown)


rdrf/rdrf/questionnaires.py
```diff
@@ -442,16 +442,8 @@ def info(self, msg):
...
-            try:
-                self._remove_mongo_data(self.registry, patient)
-                logger.info("removed dynamic data for %s for registry %s" %
-                            (patient.pk, self.registry))
-                raise PatientCreatorError("Error saving dynamic fields: %s" % ex)
-            except Exception as ex2:
-                logger.error("could not remove dynamic data for patient %s: %s" %
-                             (patient.pk, ex2))
-                raise PatientCreatorError("Error saving fields: %s. But couldn't remove bad data: %s" % (ex, ex2))
+            logger.error("Error saving dynamic data in clinical database: %s" % ex)
```
sztamas: `logger.exception` ?

rdrf/rdrf/rpc_commands.py
```diff
@@ -153,22 +153,17 @@ def rpc_update_selected_cdes_from_questionnaire(
             if len(errors) > 0:
                 raise Exception("Errors occurred during update: %s" % ",".join(errors))
     except Exception as ex:
-        should_revert = True
+        failed = True
         logger.error("Update patient failed: rolled back: %s" % ex)
 
-    if not should_revert:
+    if not failed:
```
sztamas: try/except/**else**

PR #458, opened 2017-06-27
Questionnaire Validation Improvements

id2359:
If there are required field errors, the error is scrolled to automatically.
The missing address field errors  outlined in red ( same styling as the consent validation) and scrolled to.
Instead of the red table row for required fields I've made the questionnaire use red asterisks ( as we use in the application.)

### DIFF
--- a/rdrf/rdrf/templates/rdrf_cdes/base_questionnaire.html
+++ b/rdrf/rdrf/templates/rdrf_cdes/base_questionnaire.html
@@ -47,31 +47,32 @@
         
     </head>
     <script>
-      // scroll to any validation errors
-      $(document).ready(function() {
+        $(document).ready(function() {
 
-      (function() {
-      var delay = 0;
-      var offset = 150;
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/templates/rdrf_cdes/base_questionnaire.html
```diff
@@ -30,10 +30,48 @@
...
             </style>
+
         {% endblock %}
         
     </head>
+    <script>
+      // scroll to any validation errors
+      $(document).ready(function() {
+
+      (function() {
```
sztamas: Not a biggie, but the indentation is wrong and not sure why this code is wrapped in a function?

id2359: Yeak good call - Initially I had the body outside the ready call so was thinking i'd close off the variables later I moved it into the ready call which makes it unnecessary.

PR #468, opened 2017-07-10
Validation improvements

id2359:
- Exposed schema validation as a management command
- Corrected the json schema for Modjgo objects 
- Added test of command
- Prevented form model save with a missing Section object

### DIFF
--- a/rdrf/rdrf/management/commands/check_structure.py
+++ b/rdrf/rdrf/management/commands/check_structure.py
@@ -1,0 +1,49 @@
+from django.core.management import BaseCommand
+from rdrf.models import Registry
+from rdrf.models import Modjgo
+import yaml
+import jsonschema
+
+explanation = "to do"
+
+class Command(BaseCommand):
+    help = 'Checks in clinical db against json schema(s)'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--collection',
+                            action='store',
+                            dest='collection',
+                            default="cdes",
+                            help='Collection name ( one of cdes, history or registry_specific)')
+
+    def _usage(self):
+        print(explanation)
+
+    def handle(self, *args, **options):
+        self.schemas = self._load_schema_file()
+        collection = options.get("collection", "cdes")
+
+        for modjgo_model in Modjgo.objects.filter(collection=collection):
+            data = modjgo_model.data
+            problems = self._check_for_problems(data, collection)
+
+
+    def _load_schema_file(self):
+        modjgo_schema_file = "/app/rdrf/rdrf/schemas/modjgo.yaml"
+        with open(modjgo_schema_file) as sf:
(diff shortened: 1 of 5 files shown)


rdrf/rdrf/management/commands/check_structure.py
```diff
@@ -0,0 +1,105 @@
...
+import yaml
+import jsonschema
+import errno
+import os
+
+explanation = "This command checks for schema validation errors"
+
+SCHEMA_LOCATIONS = ["/app/rdrf/rdrf/schemas/modjgo.yaml",
+                    "/env/src/django-rdrf/rdrf/rdrf/schemas/modjgo.yaml"]
+
```
sztamas: I don't like these docker container specific hardcoded paths here.
Can we do something about them?
For example get path based on a module's `__file__` or similar.

rdrf/rdrf/management/commands/check_structure.py
```diff
@@ -0,0 +1,105 @@
...
+    def handle(self, *args, **options):
+        self.test_mode = options.get("test_mode", False)
+        problem_count = 0
+        self.schema = self._load_schema()
+        registry_code = options.get("registry_code",None)
+        if registry_code is None:
+            self._print("Error: registry code required")
+            if not self.test_mode:
+                sys.exit(1)
```
sztamas: Not sure about this one, but it seems to me that you would want `sys.exit()` even if `test_mode is False`. Or does it make sense to continue at all if there is no registry code?

rdrf/rdrf/management/commands/check_structure.py
```diff
@@ -0,0 +1,105 @@
...
+        try:
+            registry_model = Registry.objects.get(code=registry_code)
+        except Registry.DoesNotExist:
+            self._print("Error: registry does not exist")
+            if not self.test_mode:
+                sys.exit(1)
```
sztamas: Same as comment at line 50.

rdrf/rdrf/management/commands/check_structure.py
id2359: Yeah I was wondering about it too ok

rdrf/rdrf/management/commands/check_structure.py
id2359: ok true

rdrf/rdrf/management/commands/check_structure.py
id2359: Actually I did this on purpose - I'll have to remember why now   ,initially I tested for /expected sys exit in the tests, i'll have to re-try it  to prod my  memory

id2359:
Ok I made the script locate the schema file without hardcoding paths

and also I've removed Test_mode and just tested for sys exit codes

PR #470, opened 2017-07-11
Preferred languages on registration

id2359:
The patient registration workflow allows choice of preferred language.

The allowed values were previously hard coded in the registration form and user admin form.

I've altered this to make both lists dynamic ( based on value of the LANGUAGES assoc list string exposed in settings.) 

Choices set on the User model's preferred language field trigger (unwanted) migrations if value is propagated from settings and a site changes the setting from puppet - so I have removed this an added a validation condition on the field instead which does not trigger a migration if the setting is changed.

I have also changed the default English.
If only English is supported, the registration form does not show a one element drop down list ( with the english being set as default)


rdrf/rdrf/registration_rdrf.py
```diff
@@ -24,4 +24,29 @@ def post(self, request, *args, **kwargs):
...
+        from django.conf import settings
+
+        # The default list is english only which we don't bother to show
+        if len(settings.LANGUAGES) == 1:
+            if settings.LANGUAGES[0][0].lower() == "en":
+                return []
+            
+        l = [] 
+
+        class LanguageWrapper:
```
sztamas: using `collections.namedtuple` would be simpler. 
I would also move this to a module (like  `models.py`) so you can re-use it in the validation code below, that currently uses slicing instead of field names.

rdrf/rdrf/registration_rdrf.py
```diff
@@ -24,4 +24,29 @@ def post(self, request, *args, **kwargs):
...
+            def __init__(self, code, name):
+                self.code = code
+                self.name = name
+
+        for code, name in settings.LANGUAGES:
+            l.append(LanguageWrapper(code, name))
+
```
sztamas: `return [LanguageWrapper(code, name) for code, name in settings.LANGUAGES]` looks more pythonic to me

rdrf/registry/groups/models.py
```diff
@@ -190,6 +188,14 @@ def can_view(self, registry_form_model):
 
         return False
 
+    def clean(self):
```
sztamas: I would put this in `clean_preferred_language` instead.

rdrf/rdrf/registration_rdrf.py
id2359: indeed

id2359:
good comments , I'll change

id2359:
I've altered the admin user form to show a drop down which is built from choices exposed in settings - this will avoid the issue with migrations being triggered, plus is better than validating free text of the code.

Issue #131, opened 2016-08-27
[RDR-1292] The list of forms in the "Fixed" context for FH is showing the FollowUp form link


ccgbuildbot:
<b><i>[reporter="lrender", created="Thu, 26 May 2016 15:21:45 +0800", resolved="Fri, 27 May 2016 13:28:58 +0800"]</i></b>
<p>This is wrong as the FollowUp form is in the other "Multiple" context form group</p>

ccgbuildbot:
<b><i>[author="lrender", created="Fri, 27 May 2016 13:28:58 +0800"]</i></b>
<p>Turns out this was not a bug - the form list was correctlly showing all forms in the context that had been created - it was the wrong context though, due to a different bug - fixing this issue has fixed this also.</p>

PR #475, opened 2017-08-08
Enhancements to base RDRF for MTM

id2359:
New features: ( bit of a grab bag)

- Patient types (MTM #131)  For MTM different forms are needed for different types of patients ( the type of patient being set at registration time). The types being: deceased patients and female carriers.

I've added a patient type field on the patient model and made the form drop down /launcher check this.
The only difference is which list of forms to display. If a patient is not given a type, the list of forms defaults to all forms - the union of "typed" forms ( ie the rest of the forms) . Precisely which forms are allowed /applicable to each patient type is set in the registry metadata json field.  ( NB this is an orthogonal  concept to an auth group being allowed to edit or view the form , which we've already implemented.) If no patient types are set, the default behaviour is all forms are applicable.

NB. The MTM client still needs to provide us with the new form definitions for these two groups.

- Reminders based on last login date. I've added a management command to detect users ( ie patients and parents ) in mtm who haven't logged for a while and send reminders using the existing email notification framework. "reminder" is now a new notification type. The mechanism to trigger the management command is django_cron. 

We still to write a trigger to hit the runcrons management command. There are timer signals in uwsgi which we cold hook into

- Added in some spanish translations for info documents.


rdrf/rdrf/management/commands/check_logins.py
```diff
@@ -0,0 +1,132 @@
...
+from django.core.management import BaseCommand
+from rdrf.models import Registry
+from rdrf.reminders import ReminderProcessor
+
+from registry.groups.models import CustomUser
+from registry.patients.models import Patient
+from registry.patients.models import ParentGuardian
+
+def send_reminder(user, registry_model,process_func=None):
+    if process_func:
```
sztamas: The conditional isn't needed if you just use:

`ReminderProcessor(user, registry_model, process_func)`

rdrf/rdrf/management/commands/check_logins.py
```diff
@@ -0,0 +1,132 @@
...
+            self._error("Registry code required")
+            sys.exit(1)
+
+        try:
+            registry_model = Registry.objects.get(code=registry_code)
+        except Registry.DoesNotExist:
+            self._error("Registry does not exist")
+            sys.exit(1)
+
+        days = options.get("days", None)
```
sztamas: `days = options.get("days", self._get_numdays(registry_model)`

rdrf/rdrf/management/commands/check_logins.py
```diff
@@ -0,0 +1,132 @@
...
+        # allow override form command line else use registry config
+        if days is None:
+            days = self._get_numdays(registry_model)
+
+        threshold = self._get_threshold(days)
+
+        action = options.get("action", None)
```
sztamas: same as
`action = options.get("action")`

rdrf/rdrf/utils.py
```diff
@@ -652,3 +655,43 @@ def get_supported_languages():
...
+    logger.debug("patient type = %s" % patient_model.patient_type)
+    patient_type_map = registry_model.metadata.get("patient_types", None)
+    logger.debug("patient type map = %s" % patient_type_map)
+    # type map looks like:
+    # { "carrier": { "name": "Female Carrier", "forms": ["CarrierForm"]} }
+    all_forms = registry_model.forms
+
+    if patient_type_map is None:
+        return all_forms
+    else:
```
sztamas: `else` not needed. Can unnest.

rdrf/rdrf/management/commands/check_logins.py
id2359: Ta - I think in some cases, this is probably due to introducing changes into the code as say an issue is noticed  ( in this case it was making something testable after the fact) and then not doing the refactoring as a result.

PR #478, opened 2017-08-11
don't initialise django app within tasks.py

aahunter:
Recently added tasks.py was triggering django initialisation, which is not what we want in prod. Initialisation of Django needs to take place in the web app.

### DIFF
--- a/rdrf/tasks.py
+++ b/rdrf/tasks.py
@@ -1,7 +1,7 @@
 """
 Define events for use with the uWSGI cron-like interface (https://uwsgi-docs.readthedocs.io/en/latest/Cron.html)
 
-Care must be taken with Django imports. We want application initiliasation to be triggered by the web application
+Care must be taken with Django imports. We want application initialisation to be triggered by the web application
 not by this file. Hence we are checking to see if the django app is ready before performing some imports.
 
 """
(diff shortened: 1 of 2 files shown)


rdrf/tasks.py
```diff
@@ -1,32 +1,31 @@
 """
 Define events for use with the uWSGI cron-like interface (https://uwsgi-docs.readthedocs.io/en/latest/Cron.html)
 
+Care must be taken with Django imports. We want application initiliasation to be triggered by the web application
```
sztamas: Spelling error: initialisation

rdrf/tasks.py
```diff
@@ -1,32 +1,31 @@
...
-    def debug_timer(num):
-        logger.debug("")
-
-
 @timer(300)
 def django_cron_heartbeat(num):
     """
     A heartbeat for django-cron (https://github.com/Tivix/django-cron)
     """
+    if apps.apps_ready is not True:
```
sztamas: `if not apps.apps_ready:`

rdrf/tasks.py
```diff
@@ -1,32 +1,31 @@
...
+        # Django is not initialised, the logger config in settings will not be active
+        print("Deferring calling django-cron heartbeat until django app is ready")
```
sztamas: Why not use logging instead of `print`?

aahunter: Because the app is not initialised and therefore there is no logging config.

PR #482, opened 2017-08-17
Modifiying link order for follow up form

schangccg:


rdrf/rdrf/wizard.py
```diff
@@ -77,10 +78,33 @@ def _multiple_form_groups(self):
...
+
+            try:
+                return self.patient_model.get_form_value(self.registry_model.code,
+                                                         form_name,
+                                                         section_code,
+                                                         cde_code,
+                                                         multisection=is_multisection,
+                                                         context_id=context_model.id)
+            except:
+                 raise Exception("Could not get CDE data (Date of assessment).")
```
id2359: On thinking about it it's probably best not to check from exception here. Also we can't assume in the message that date of assessment is being used ( it really depends on what cde_code is and that is set at design time when the ContextFormGroup model is created

id2359: Also I usually use 
```python
<repo name> #<ticketnumber> 
```
in my commit messages because we have several projects which use rdrf as a framework

schangccg: Ok will make a note of that for future commit messages.

PR #495, opened 2017-08-28
Security fix

id2359:
Editing or viewing a patient data (consent, demographic, clinical) record can only occur on the precondition that the user and patient are in the correct security relationship which is:

A user is the   patient administering their own data
A user is the parent or guardian of the patient
A user is a staff user ( non patient) and is in the same working  group

### DIFF
--- a/rdrf/rdrf/patient_view.py
+++ b/rdrf/rdrf/patient_view.py
@@ -507,9 +507,7 @@ class PatientEditView(View):
             patient_id, registry_code, request)
 
         if not check_patient_user(request.user, patient):
-            logger.info("User %s not authorised to view patient %s" % (request.user,
-                                                                       patient))
-            return HttpResponseRedirect("/")
+            raise PermissionDenied()
         
         registry_model = Registry.objects.get(code=registry_code)
 
@@ -552,6 +550,10 @@ class PatientEditView(View):
     def post(self, request, registry_code, patient_id):
         user = request.user
         patient = Patient.objects.get(id=patient_id)
+
+        if not check_patient_user(user, patient):
+            raise PermissionDenied()
+
         patient_relatives_forms = None
         actions = []
 
--- a/rdrf/rdrf/patient_view.py
+++ b/rdrf/rdrf/patient_view.py
@@ -30,6 +30,9 @@ from rdrf.wizard import NavigationWizard, NavigationFormType
 from rdrf.components import RDRFContextLauncherComponent
 from rdrf.components import RDRFPatientInfoComponent
 
+from rdrf.security_checks import check_patient_user
+
+
 import logging
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


rdrf/rdrf/patient_view.py
```diff
@@ -30,6 +30,10 @@
 from rdrf.components import RDRFContextLauncherComponent
 from rdrf.components import RDRFPatientInfoComponent
 
+from rdrf.security_checks import security_check_user_patient
+from django.core.exceptions import PermissionDenied
```
aahunter: Minor - PermissionDenied import may be unused.

rdrf/rdrf/security_checks.py
```diff
@@ -0,0 +1,66 @@
...
+                                                           patient_model.pk))
+    raise PermissionDenied()
+
+def security_check_user_patient(user, patient):
+    # either user is allowed to act on this record ( return True)
+    # or not ( raise PermissionDenied error)
+    if user.is_superuser:
+        return True
+
+    if patient is None:
```
aahunter: Can we remove this special case

rdrf/rdrf/security_checks.py
```diff
@@ -0,0 +1,66 @@
...
+        return True
+
+    if type(patient) is int:
+        patient_model = Patient.objects.get(id=patient_id)
+    elif type(patient) is str:
+        patient_model = Patient.objects.get(id=int(patient_id))
+    else:
+        patient_model = patient
```
aahunter: Can we make it a pre-condition that a model is passed?

rdrf/rdrf/security_checks.py
```diff
@@ -0,0 +1,66 @@
...
+    except:
+        return False
+
+def _user_is_patient_type(user):
+    return any([_get_prop(user, "is_patient"),
+                _get_prop(user, "is_parent"),
+                _get_prop(user, "is_carrier")])
+
+
+def _security_violation(user, patient_model):
```
aahunter: Call this _raise_security_violation to make the code below more obvious where the function exits.

rdrf/rdrf/security_checks.py
id2359: yes I thought it better also after thinking

rdrf/rdrf/security_checks.py
```diff
@@ -0,0 +1,66 @@
...
+        return True
+
+    if type(patient) is int:
+        patient_model = Patient.objects.get(id=patient_id)
+    elif type(patient) is str:
+        patient_model = Patient.objects.get(id=int(patient_id))
+    else:
+        patient_model = patient
+    
+    if _user_is_patient_type(user):
```
aahunter: Perhaps remove on level of indentation (else can be redundant) and always raise at the end.

aahunter:
Reviewed, subject to tests...

rdrf/rdrf/security_checks.py
id2359: nod - done

PR #513, opened 2017-09-13
Clean up modjgo refactoring and patient listing tweaks

id2359:
I started this branch and called it "experimental" - which was a bad name - I've created this branch from that and done some work based on the issues raised in my issues document in google drive in it:

- removed lots of logging code
- removed the (dead) Adjudication code 
- renamed Modjgo to ClinicalData
- migration to add new fields to ClinicalData + a data migration to extract the current data form the json and populate these fields.
- put some ids on the ClinicalData model which were previously embedded inside the json data field
(I've left them embedded in there for now also )
- refactored the constructor for ClincialData and the find method on its Model Manager
- unbroke a couple of issues caused by this ( FollowUp forms in FH )
- fixed and recreated the test zips ( the exported test data had Modjgo names so I edited them and rezipped) then recreated them using develop.sh.
- tweaked some code in the patient listing to slightly increase the speed ( prefetch some objects etc)

For the patient listing the actual rate determining step is the forms button code generation - this is will be refactored in another branch


rdrf/rdrf/dynamic_data.py
```diff
@@ -486,7 +474,7 @@ def load_dynamic_data(self, registry, collection_name, flattened=True):
     def get_cde_val(self, registry_code, form_name, section_code, cde_code, collection="cdes"):
         modjgo_queryset = self._get_record(registry_code, collection)
```
aahunter: We should rename these modjgo_ variables to clinical_data_ or the like.

rdrf/rdrf/dynamic_data.py
```diff
@@ -714,6 +692,9 @@ def save_dynamic_data(self, registry, collection_name, form_data, multisection=F
             # create context_model NOW  to get context_id
             # CREATE MODE is used ONLY by multiple context form groups to enable cancellation in GUI
             context_id = self._create_context_model_on_fly()
+            # we've refactored the ClinicalData object so that context_id is now on the model:
+            record.context_id = context_id
+            # keeping this line for backward compatibility for now
```
aahunter: What does "backward compatibility" mean in this context?

rdrf/rdrf/migrations/0040_mongo_django.py
```diff
@@ -9,7 +9,9 @@
     mongo_django = undjango_mongo = None
 
 def copy_collections(apps, schema_editor):
+    print("in copy_colections")
```
aahunter: Do you want to keep this debug code in the migrations?

rdrf/rdrf/migrations/0056_auto_20170911_1039.py
```diff
@@ -0,0 +1,28 @@
...
+    try:
+        ClinicalData = apps.get_model("rdrf", "ClinicalData")
+        for clinical_data in ClinicalData.objects.all():
+            clinical_data.django_model = clinical_data.data.get("django_model", "")
+            clinical_data.django_id = clinical_data.data.get("django_id", 0)
+            clinical_data.context_id = clinical_data.data.get("context_id", None)
+            clinical_data.save()
+    except ProgrammingError:
+        # thrown when run on wrong db
+        print("not running clinical data update on main db")
```
aahunter: Should this fail the migration?

aahunter:
Added some inline comments. Held back on more "remove mongo references" suggestions as that is tied up in a larger task.

aahunter:
oh and LGTM

rdrf/rdrf/migrations/0056_auto_20170911_1039.py
id2359: We had as issue like this before - the database router is meant to route any operations on a model to the right db - but for the data migration it runs on both despite the model being listed in the router as residing on the clinical db. By catching this error that was thrown , I allow the migrations to continue where this one gets run on the correct db, rather than halt the whole process.

rdrf/rdrf/migrations/0040_mongo_django.py
id2359: No, I over looked this, I'll remove.

rdrf/rdrf/dynamic_data.py
id2359: Good question - I think I was thinking about reports - I'll have a look at them now to check

PR #515, opened 2017-09-14
Login trouble self assistance feature

sztamas:
As discussed in muccg/rdrf-ccg#940


rdrf/rdrf/templates/registration/login_assistance_account_not_locked_email.html
```diff
@@ -0,0 +1,23 @@
+{% load i18n %}{% autoescape off %}
+Hello  Mr/Ms {{user.last_name}},
+
+{% blocktrans %}You're receiving this email because you were having trouble signing in and requested help.{% endblocktrans %}
+
+{% blocktrans %}Your account is NOT locked. You should be able to log in if you're providing the correct username and password.{% endblocktrans %}
```
aahunter: Minor comment - I don't think we need to tell the user the account is not locked.

rdrf/rdrf/templates/registration/login_assistance_email.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}{% autoescape off %}
+Hello  Mr/Ms {{user.last_name}},
+
+{% blocktrans %}You're receiving this email because you were having trouble signing in and requested help.{% endblocktrans %}
+
+{% blocktrans %}Your account has been locked for security purposes and the system needs to reverify you.{% endblocktrans %}
```
aahunter: I don't think we need to mention account locking here. Perhaps change to:

"The system needs to verify your account."

rdrf/rdrf/templates/registration/login_assistance_form.html
```diff
@@ -0,0 +1,20 @@
+{% extends "rdrf_cdes/base.html" %}
+{% load i18n %}
+
+{% block content %}
+
+<br>
+
+<div class="alert alert-warning">
+  {% trans "Trouble signing in? Did you forgot your password or your account has been locked for security reasons? Enter your email address below, and we'll email you instructions on what to do." %}
```
aahunter: I don't think we need to mention account lock here.

"Trouble signing in? Did you forgot your username or password? Enter your email address below, and we'll email you instructions on what to do."

rdrf/rdrf/templates/registration/login_assistance_sent.html
```diff
@@ -0,0 +1,15 @@
+{% extends "rdrf_cdes/base.html" %}
+{% load i18n %}
+
+{% block content %}
+<br>
+
+<div class="alert alert-info">
+{% trans "We've emailed you instructions for unlocking your account, if an account exists with the email you entered. You should receive them shortly." %}
```
aahunter: Can this become:

"If an account exists with the email you entered we will email you instructions to verify your account, . You should receive the email shortly."

rdrf/rdrf/templates/registration/password_change_form.html
```diff
@@ -0,0 +1,37 @@
+{% extends "rdrf_cdes/base.html" %}
+{% load i18n %}
+{% load static from staticfiles %}
+
+{% block content %}
+
+<div class="row">
+    <div class="col-md-6 col-md-offset-3">
+        <div class="alert alert-info">
+            {% trans "Please enter your old password, for security's sake, and then enter your new password twice so we can verify you typed it in correctly." %}
```
aahunter: Edit to text:

"Please enter your old password, and then enter your new password twice so we can verify you typed it in correctly."

rdrf/rdrf/auth/forms.py
```diff
@@ -0,0 +1,209 @@
...
+             use_https=False, token_generator=default_token_generator,
+             from_email=None, request=None, html_email_template_name=None,
+             extra_email_context=None):
+
+        email = self.cleaned_data["email"]
+        context = self._common_context(request=request,
+                domain_override=domain_override, use_https=use_https, extra_email_context=extra_email_context)
+
+        def _user_is_nonprivileged(user):
+            return user.is_patient or user.is_parent or user.is_carrier
```
id2359: good

rdrf/rdrf/templates/registration/password_change_form.html
```diff
@@ -0,0 +1,37 @@
...
+        </div>
+        <form action="" method="post" class="form" id="new-password-form">
+            {% csrf_token %}
+            {{ form.non_field_errors }}
+            <div class="form-group">
+                <label for="old_password">Old password:</label>
```
aahunter: Change text:

Current password:

rdrf/rdrf/auth/views.py
```diff
@@ -0,0 +1,175 @@
...
+            _('Your account is suspended because your password expired.'),
+        UserDeactivation.TOO_MANY_FAILED_LOGINS:
+            _('Your account is suspended because you (or someone else) entered an incorrect password too many times.')
+    }
+    reason = REASON_TEXTS.get(deactivation_reason, DEFAULT_REASON_TEXT)
+
+    def is_password_change_required(user):
+        # If the password has expired always make them change their password or they will
+        # be locked out when they try to log in next time.
+        return deactivation_reason == UserDeactivation.PASSWORD_EXPIRED or is_password_expired(user)
```
id2359: could there be a userdeactivation object with reason UserDeactivation.PASSWORD_EXPIRED but is_password_expired(user)  false?

rdrf/rdrf/templates/registration/login_assistance_can_not_verify.html
```diff
@@ -0,0 +1,14 @@
+{% extends "rdrf_cdes/base.html" %}
+{% load i18n %}
+{% load static from staticfiles %}
+
+{% block content %}
+<div class="row">
+    <div class="col-md-6 col-md-offset-3">
+        <div class="alert alert-warning">
+            {% trans "Unfortunately we don't have the required information to verify your identity. Please contact the registry owners for assistance." %}
```
id2359: Presumably a patient, say, would have email of the owners or would have enough info in initial registration emails to locate the splashcreen (?) with that info. But we should check.

rdrf/rdrf/templates/registration/login_assistance_form.html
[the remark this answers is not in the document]
sztamas: The account locking terminology has been dropped as suggested by Adam, so this doesn't apply anymore.

rdrf/rdrf/auth/views.py
sztamas: Yes. `UserDeactivation.PASSWORD_EXPIRY` is set by `django-useraudit` only if `django-useraudit` deactivated the user account and because of password expiry.

The `is_password_expired` check is there for cases when the user has been deactivated for other reasons, but their password is also expired.

If we remove that check and allow them to reactivate their account without changing their password, they will get locked out because of their password expired when they will try to log in.

sztamas:
Changed all the text as suggested by Adam to avoid the account locking terminology.

id2359:
I've tested this locally Tamas - looks good-  - the only issue I found was this:

1. I created a "patient" who locked themselves out with too many failed password attempts: received the verification link email. OK

2. As an admin I then checked the self unlock prevention checkbox in admin interface for this user. 

3. The patient user could still "reactivate" through the verify account workflow.

4 However ,  If the prevent unlock checkbox is checked before the password failures, the having trouble email sent does not contain a verifcation link

sztamas:
OK, @id2359 I've just added an extra check for that use case.

PR #516, opened 2017-09-15
Ajax patient listing

id2359:
Current patient listing is very slow ( most noticeably in FH). The main cause is the retrieval of the forms list dropdown for each patient on the page. I have moved this logic into an ajax call so the forms list is rendered on demand. This has reduced  the page load time by a factor 11-12 in my testing with a db of 50000 dummy patients.

The forms list itself has been enhanced slightly: Currenly for FH, say, all folllow ups are shown in a list. The new implementation shows the latest 10. Also an "Add Follow Up" link has been appended to the list allowing easy updating of a patient record resulting from a search. I've toned down the colour to make it less garish also.

### DIFF
--- a/rdrf/rdrf/components.py
+++ b/rdrf/rdrf/components.py
@@ -422,3 +422,46 @@ class FormsButton(RDRFComponent):
                 return self.context_form_group.forms[0].nice_name + "s"
             else:
                 return self.context_form_group.name
+
+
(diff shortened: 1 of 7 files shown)


aahunter:
For the record, we discussed verbally and I neglected to put comments inline. PR LGTM

Issue #525, opened 2017-10-03
Add two-factor authentication


sztamas:
Virtual MFA.
Using 2FA is optional, suggested for privileged users. New flag on user that allows 2FA as a requirement.

sztamas:
Just to sum  up the discussions on the PR's code review...

We decided to keep the "Disable two-factor auth" menu item even for users that are required to have two-factor auth.

When clicked they will be presented with a warning saying that they are required to have 2fa, so they will not be able to use the system until they set up 2fa again.

If they go on with the disabling they will have to set up 2fa again, but this gives them a way to set up a new device for 2fa without requiring the assistance of an admin user.

Use case: allow user to change their 2fa device.
User signs in to the system with password and current device token.
User disables 2fa by using the "Disable two-factor auth" menu.
User is taken to the "Set up two-factor authentication" page and sets up the new device.

PR #526, opened 2017-10-03
Add two-factor authentication #525

sztamas:


rdrf/rdrf/auth/views.py
```diff
@@ -6,32 +6,65 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+@receiver(user_logged_in)
+def user_login_callback(sender, request=None, user=None, **kwargs):
+    if is_user_privileged(user) and default_device(user) is None:
+        link = ('<a href="%(url)s" class="alert-link">' + _('click here') + '</a>') % {'url': reverse('two_factor:setup')}
+        msg = mark_safe(
+                _('We strongly recommend that you protect your account with Two-Factor authentication. '
```
id2359: So this message will always appear as a nag ?

sztamas: Yes, but only for "system users".

id2359: I found that if as an admin I made a user use 2fa and then logged in as that user I get the 2fa setup screen as expected, but after I set up the 2fa ,  am still greeted by that 2fa recommendation message ( despite having just set it up)

id2359: Also after being "forced" to setup 2fa by an admin, the disable 2fa option still exists in the drop down, which I can then click. Then when I log back in I'm forced to set it up again.

So if 2fa is being made mandatory ( via the admin setting it as required in the admin interface), shouldn't the cancel option simply not be presented to the user at all in this case?

rdrf/rdrf/auth/views.py
```diff
@@ -40,7 +40,7 @@
 
 @receiver(user_logged_in)
 def user_login_callback(sender, request=None, user=None, **kwargs):
-    if is_user_privileged(user) and default_device(user) is None:
+    if is_user_privileged(user) and not user.require_2_fact_auth and default_device(user) is None:
```
id2359: looks good

Issue #191, opened 2016-08-27
[RDR-1335] DM1- Questionnaire Responses- Link to patient Info to open in new tab (after loading existing patient)


ccgbuildbot:
<b><i>[reporter="knapier", created="Mon, 20 Jun 2016 11:36:07 +0800"]</i></b>

PR #535, opened 2017-10-16
Fix form navigation chevrons not wrapping around correctly mtm #191

schangccg:


schangccg:
MTM Ticket: https://github.com/muccg/mtm/issues/191

id2359:
Approved - will merge into branch mtm-phase2 so I can aggregate all changes and test prior to merging into next_release

PR #536, opened 2017-10-17
Removed Mongo related deps,containers,data migs and scripts

id2359:


aahunter:
Perhaps also remove some legacy scripts in:

```
rdrf/rdrf/scripts/
rdrf/rdrf/management/commands/
```

PR #543, opened 2017-11-03
Changes for MTM in base rdrf

id2359:
- Default patient type forms list is now specified statically in the registry metadata.
- Translation files recreated using improved script
-  Wording changes


id2359:
Given go ahead to merge will do now

PR #559, opened 2018-01-23
Clinician workflow

id2359:
Branch contains:

- A new clinician view and form.
- A new mechanism to associate patient consent with behaviours in the system ( via "Consent Rules".)
- Triggering of a notification when a user nominates a clinician not in the list of clinician users in the system.
- Restriction of the patient listing for clinicians to their own patients that have consented.


rdrf/rdrf/clinician_view.py
```diff
@@ -0,0 +1,260 @@
...
+    def _get_objects(self, request, registry_code, patient_id):
+        self.request = request
+        self.user = request.user
+        self.message = None
+        if self.user.is_parent:
+            self.parent = ParentGuardian.objects.get(user=self.user)
+        else:
+            self.parent = None
+        self.patient_model = get_object_or_404(Patient, pk=patient_id)
+        if not security_check_user_patient(self.user, self.patient_model):
```
grahame: I think if this function checks and enforces permissions it should be a bit clearer; or could this check be broken out and put somewhere more obvious? just so that someone doesn't lose it refactoring later on

rdrf/rdrf/clinician_view.py
```diff
@@ -0,0 +1,260 @@
...
+        self._get_objects(request, registry_code, patient_id)
+
+        if self.clinician_form.is_valid():
+            other_clinician_model = self.clinician_form.save()
+            logger.debug(
+                "saved other clinician with pk %s" %
+                other_clinician_model.pk)
+            if other_clinician_model.user:
+                other_clinician_model.patient.clinician = other_clinician_model.user
+                other_clinician_model.patient.save()
```
grahame: should we bundle these calls to save() in a transaction so we don't end up with stuff dangling if one fails?

rdrf/registry/patients/models.py
```diff
@@ -1008,14 +1008,67 @@ def family(self):
...
-    clinician_address = models.CharField(max_length=200, null=True)
+    clinician_name = models.CharField(max_length=200, blank=True,null=True)
+    clinician_hospital = models.CharField(max_length=200, blank=True, null=True)
+    clinician_address = models.CharField(max_length=200, blank=True, null=True)
     clinician_email = models.EmailField(max_length=254, null=True, blank=True)
     clinician_phone_number = models.CharField(max_length=254, null=True, blank=True)
+    user = models.ForeignKey(CustomUser, blank=True, null=True)
+
+    def synchronise_working_group(self):
+        if self.user:
```
grahame: I'd probably flip this `if` and return from the function if self.user is None, then you can drop the indent and anyone adding code won't accidentally create a security issue

rdrf/registry/patients/models.py
```diff
@@ -1008,14 +1008,67 @@ def family(self):
...
+                else:
+                    logger.debug("can't synch - patient not set")
+            else:
+                logger.debug("can't synch - user_other is True", None)
 
 
+@receiver(post_save, sender=ClinicianOther)
+def other_clinician_post_save(sender, instance, created, raw, using, update_fields, **kwargs):
+
+    if not instance.user and instance.use_other:
```
grahame: same thing here; I'd invert the `if`

rdrf/rdrf/clinician_view.py
id2359: Good point - been thinking of looking at transaction management across the app ( for registration also)

rdrf/registry/patients/models.py
id2359: will do

id2359:
I've merged and will recreate test zips locally

Issue #425, opened 2017-03-09
Consent page- upload consent file: deletion of multisection deletes uploaded file but does not delete section


knapier:
Saving of the page after deleting the mutiection says 'pateint saved successfully', but red error message appears below 'deleted' multisection (that has the uploaded file deleted but multisection remains).

* Select a valid choice. That choice is not one of the available choices.

Affects 2.0.1-2.0.3

id2359:
Replicated on FH
Add 1 file save f1
Add another file and save f2
links for both files are visible only after navigating from the page and back again ( to trigger the get view)
mark f1 for deletion 
the "section" is deleted but the mark for deletion checkbox shows and is checked

knapier:
Note, was introduced in v2 (testing on v1.7.14 works as expected).

id2359:
We caused this by removing  a redirect to the get view ( which was preventing a success message from being displayed to the user ) - the originating  POST view does not build the response properly 

Looking into using SuccessMessageMixin

id2359:
Fix just needed to put back redirect and add call  messages api


Fixed #426 also

PR #564, opened 2018-01-29
accommodate different project logo width (fkrp #425)

schangccg:
for ticket: https://github.com/muccg/fkrp/issues/425


schangccg:
for ticket: https://github.com/muccg/fkrp/issues/425

Issue #529, opened 2017-10-05
Implement password validation when creating new user


knapier:
Password validation is not enforced when you create a new user and set up their password.

'Menu', 'Users', 'add'.

Password validation is enforced when you change the users password after creation of the user (registry specific). For example:

Your password must contain at least 8 characters.
Your password can't be a commonly used password.
Your password must contain at least 1 uppercase letter.
Your password must contain at least 1 lowercase letter.
Your password must contain at least 1 number.
Your password must contain at least 1 special character.

schangccg:
pr: https://github.com/muccg/rdrf/pull/572

schangccg:
Implemented.

PR #572, opened 2018-02-09
add password validation

schangccg:
for ticket: https://github.com/muccg/rdrf/issues/529


rdrf/registry/groups/admin_forms.py
```diff
@@ -63,13 +64,14 @@ class Meta:
         model = get_user_model()
         fields = ('email',)
 
+
     def clean_password2(self):
-        password1 = self.cleaned_data.get("password1")
         password2 = self.cleaned_data.get("password2")
-        if password1 and password2 and password1 != password2:
-            raise forms.ValidationError("Passwords don't match")
+
```
id2359: lint

PR #575, opened 2018-02-13
change 2fa cancel link

schangccg:
for ticket: https://github.com/muccg/rdrf-ccg/issues/988

### DIFF
--- a/rdrf/rdrf/templates/two_factor/_wizard_actions.html
+++ b/rdrf/rdrf/templates/two_factor/_wizard_actions.html
@@ -1,9 +1,22 @@
 {% load i18n %}
 
-{% if cancel_url %}
-  <a href="{{ cancel_url }}"
-     class="btn btn-link pull-right">{% trans "Cancel" %}</a>
+
+{% if request.user.is_authenticated %}
+
+    {% if request.user.registry_code %}
+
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/templates/two_factor/_wizard_actions.html
```diff
@@ -1,10 +1,23 @@
...
+            <a href="{% url 'parent_page' request.user.registry_code %}" class="btn btn-link pull-right">{% trans "Cancel" %}</a>
+        {% elif request.user.is_patient %}
+            <a href="{% url 'patient_page' request.user.registry_code %}" class="btn btn-link pull-right">{% trans "Cancel" %}</a>
+        {% else %}
+            <a href="{% url 'patientslisting' %}" class="btn btn-link pull-right">{% trans "Cancel" %}</a>
+        {% endif %}
+
+    {% endif %}
+
 {% endif %}
```
id2359: check behaviour of cancel link on base rdrf say the FH registry as this does not have a "home" page

schangccg: Included case for RDRF/FH where there is no home page. The user would be returned to the main page (old behaviour).

PR #598, opened 2018-05-04
Clinician notification

id2359:
When a patient selects a clinician as their clinician , the clinician user is now notified via our existing notification framework.


id2359:
merged

PR #607, opened 2018-05-29
Add 2fa popup

ccgdev004:
Ticket: https://github.com/muccg/mtm/issues/300


id2359:
Just ensure that when you update the po file that the msg ids of the popup message remain together in the po file to make it easier for the translators.

PR #622, opened 2018-08-10
changed exception to log message

krupalip:
will now skip with log if no source/target section found in clinicaldata form


krupalip:
@id2359  could you please review it? Thanks Krupali

Issue #590, opened 2018-04-04
Use  Reversion on ClinicalData model


id2359:

id2359:
The ClinicalData object holds all clinical information (PHI) ( form data, progress data, longitudinal data) - by linking to django reversion we can hold history of user changes at the record level that is automatically inspectable/auditable via the admin interface.

id2359:
see rdrf.models.definition.models.ClinicalData

PR #637, opened 2018-10-08
rdrf #590 added django reversion on ClinicalData model

aminur80:
Adding reversion module in our application.
Apply reversion on ClinicalData model so that the model can have history of changes. This enables the Admin user to revert back to any older version.
The reversion for ClinicalData is applied at Admin level.


rdrf/setup.py
```diff
@@ -47,6 +47,7 @@
     "django-two-factor-auth==1.6.2",
     "phonenumberslite==8.8.1",
     "qrcode==4.0.4",
+    "django-reversion",
```
id2359: pin to version number

PR #671, opened 2018-11-19
allow user to send patient email or generate QRCode

aminur80:
- Allow the user from Clinical system to send email or generate QRCode for patient while creating a survey request
- By default it select QRcode from dropdown


rdrf/rdrf/models/proms/models.py
```diff
@@ -159,31 +166,34 @@ class SurveyRequest(models.Model):
...
-                return False
+            if (self.communication_type == 'qrcode'):
+                try:
+                    self._send_proms_request()
+                    logger.debug("sent request to PROMS system OK")
+
+                except PromsRequestError as pre:
+                    logger.error("Error sending survey request %s: %s" % (self.pk,
+                                                                        pre))
+                    self.log_error(pre)
```
id2359: Conventionally in Python , methods which are intended to be used internally ( not as part of the public interface of a class) are named like this _some_method_name

PR #677, opened 2018-11-21
consent not considered for the calculation of  progress bar

aminur80:
- Consent page is not considered for calculating the progress
- Progress bar is removed from Consent page

### DIFF
--- a/rdrf/rdrf/frontend/src/app/index.tsx
+++ b/rdrf/rdrf/frontend/src/app/index.tsx
@@ -41,12 +41,14 @@ class App extends React.Component<AppInterface, object> {
         return this.props.stage == 0;
     }
 
+
     getProgress(): number {
         let numQuestions: number = this.props.questions.length;
-        let numAnswers: number = Object.keys(this.props.answers).length;
-        if (this.atEnd() && this.isQuestionAnswered()) {
-            numAnswers = numAnswers - 1;
-        }
+        let consentQuestionCode = this.props.questions[numQuestions - 1].cde;
+        let allAnswers = Object.keys(this.props.answers).filter(val => {
+            return val != consentQuestionCode;
+        });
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/frontend/src/app/index.tsx
```diff
@@ -44,6 +44,10 @@ class App extends React.Component<AppInterface, object> {
     getProgress(): number {
         let numQuestions: number = this.props.questions.length;
         let numAnswers: number = Object.keys(this.props.answers).length;
+        if (this.atEnd() && numQuestions==numAnswers) {
```
id2359: If the progress is not rendered at the end, this.atEnd() will always be false

rdrf/rdrf/frontend/src/app/index.tsx
```diff
@@ -44,6 +44,10 @@ class App extends React.Component<AppInterface, object> {
     getProgress(): number {
         let numQuestions: number = this.props.questions.length;
         let numAnswers: number = Object.keys(this.props.answers).length;
```
id2359: don't include consent question as an answer - filter out wth lodash?

rdrf/rdrf/frontend/src/app/index.tsx
```diff
@@ -41,12 +41,14 @@ class App extends React.Component<AppInterface, object> {
...
 
+
     getProgress(): number {
         let numQuestions: number = this.props.questions.length;
-        let numAnswers: number = Object.keys(this.props.answers).length;
-        if (this.atEnd() && this.isQuestionAnswered()) {
-            numAnswers = numAnswers - 1;
-        }
+        let consentQuestionCode = this.props.questions[numQuestions - 1].cde;
+        let allAnswers = Object.keys(this.props.answers).filter(val => {
```
id2359: cool

Issue #678, opened 2018-11-22
Make survey questions non mandatory


aminur80:
Currently the Questions in the survey need to be answered, otherwise user cannot proceed to the next one. The user should be allowed not to answer a question and proceed to the next one.

NB. This was the original behaviour and users have flipped back so it should be matter of removing the checks we added.

id2359:
Aminur suggested this link: This might be a useful reading - https://www.surveymonkey.com/curiosity/make-survey-questions-required/

PR #685, opened 2018-11-23
Make survey questions non mandatory

aminur80:
- Make survey question non mandatory (initial commit + rdrf #678 make questions non mandatory)
- Fix - Survey request page shows broken text and underline (rdrf #678 make questions non mandatory)
- Modify the complete survey page (commit - modify the survey complete page)

Some changes in email_notification is due to merge 

It would make more sense if I squash all commits into single one. That need to be tested.

### DIFF
--- a/rdrf/rdrf/templates/proms/proms_completed.html
+++ b/rdrf/rdrf/templates/proms/proms_completed.html
@@ -1,7 +1,20 @@
-<html>
-    <body>
-
-    <h1>You have completed the survey - thank you!</h1>
-    </body>
-
+{% load static from staticfiles %}
+<!DOCTYPE html>
+<html lang="en">
+   <head>
+      <title>Thank you</title>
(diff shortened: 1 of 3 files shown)


rdrf/rdrf/templates/proms/proms_completed.html
```diff
@@ -1,7 +1,20 @@
...
-</html>
+{% load static from staticfiles %}
+<!DOCTYPE html>
+<html lang="en">
+   <head>
+      <title>Thank you</title>
+      <meta name="viewport" content="width=device-width, initial-scale=1">
+      <link rel="stylesheet" href="{% static 'bootstrap-3.3.6/css/bootstrap.min.css' %}">
+   </head>
+   <body>
```
id2359: Please discuss any arbritary ( non-ticketed changes first) - we need to create separate tickets for all work and prioritise

PR #699, opened 2018-12-04
patient survey consent is saved as consent answer

aminur80:
- configure cde_code and consent_question code as part of registry metadata
- Consent Question need to be created
- PROMS survey - consent result is saved as Registry Consent Value


rdrf/rdrf/services/rest/views/proms_api.py
```diff
@@ -149,6 +149,9 @@ def _update_proms_fields(self, survey_request, survey_data):
         # clinical system
         logger.debug("updating downloaded proms for survey request %s" % survey_request.pk)
         patient_model = survey_request.patient
+        consents = [(c["cde_code"], c["consent_code"]) for c in self.registry_model.metadata.get("consents", [])]
```
id2359: Instead of an association list ( list of pairs / 2 element list )   )

use a dictionary e.g.

d = { cde_code1:   consent_code1,
        cde_code2:   consent_code2,
   etc }


instead if  iterating

consent_question_code = d.get(cde_code, None)

if consent_question_code is not None:
     # this means is a consent ....

rdrf/rdrf/services/rest/views/proms_api.py
```diff
@@ -162,11 +164,14 @@ def _update_proms_fields(self, survey_request, survey_data):
...
-                for cde_proms_code, consent_code in consents :
-                    if cde_code == cde_proms_code:
-                        self._update_consentvalue(patient_model, consent_code, value)
-                        is_consent = True
-                        break
+                logger.debug("cde code  = %s" % cde_code)
+                consent_question_code = consent_dict.get(cde_code, None)
+                logger.debug("Consent Question Code X %s" % consent_question_code)
+                if consent_question_code is not None:
+                     consent_code = consent_dict[cde_code]
```
id2359: Use consent_question_code

PR #714, opened 2018-12-20
Feature/upgrade django

aminur80:
Upgrade to Django 2.1.4


aminur80:
- Update ForeignKey
- Remove urlResolvers
- Resolve AttributeError at django.contrib.auth.view
- Replace Library.assignment_tag with Library.simple_tag
- Replace MIDDLEWARE_CLASSES with MIDDLEWARE
- Reversing Admin Urls
- TypeError: Remove direct assignment to the forward side of a manay-to-many set
- TypeError when context must be a dict so have to use context.flatten()
- Use app_name='some name' at urls.py and use the app_name in the {% url %}
- Example: app_name='proms' and use in {% url proms:survey_endpoint %}
- Add renderer at render() function

id2359:
Please check the default on_delete behaviour on  foreign keys in the old version and preserve

PR #715, opened 2019-01-03
Hide add context button for Baseline Proms

aminur80:
- add allow_proms_add string at registry metadata and set the value: True or False
- Based on the value at the string, the "Add" button at the Baseline Proms is either visible or not.


rdrf/rdrf/forms/components.py
```diff
@@ -135,10 +135,19 @@ def _get_template_data(self):
...
+            "allow_proms_add": self._get_proms_add(),
         }
 
         return data
 
+    def _get_proms_add(self):
+        metadata = self.registry_model.metadata
+        allow_proms_add = metadata.get("allow_proms_add", None)
+        if allow_proms_add is None:
+            return ""
```
id2359: Would make more sense for this to return False - returning empty string relies on python's truthyness evaluation rules and will work but why not return an explicit Boolean?

rdrf/rdrf/forms/components.py
```diff
@@ -135,10 +135,19 @@ def _get_template_data(self):
...
+        else:
+            return allow_proms_add in ['True', 'true', '1', 'yes', 'Yes']
```
id2359: Just specify that we use true or false

You could also just use a string in the features list and check registry_model.has_feature

features: [ ... ,  "can_add_proms"]


reg.has_feature("can_add_proms")


pref stick to this pattern

PR #750, opened 2019-01-25
Slider blackbox

aminur80:
- adding blackbox along with slider

### DIFF
--- a/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
+++ b/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
@@ -3,7 +3,7 @@ import * as _ from 'lodash';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
-import { Form, FormGroup, Label, Input, Col, Row } from 'reactstrap';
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
```diff
@@ -101,13 +103,21 @@ class Question extends React.Component<QuestionInterface, object> {
                 </FormGroup>
                 {
                     (question.spec.tag=='integer' ?
-                        <div style={style}>
-                            <Slider vertical min={question.spec.min}
+
+                        <div className='row'>
+                            <div className="col">
+                                <div className="float-right" style={box_style}>
+                                    <p className="text-center" style={p_style}>YOUR HEALTH RATE TODAY <b>{this.props.answers[question.cde]}</b></p>
```
id2359: We'll have to modify this so that a component is rendered based on the spec in a more fine grained way
We are assuming that every field in the survey which is of integer type is a health rating , which is obviously wrong.
Some thing like:
```
widgetType = HealthRating | ...

let widget = getWidgetFromQuestion(question)
then

switch(widget) {
      case Widgets.HEALTH_RATING:
               let react_component = <HealthRating
      case ...
               let ..

PR #760, opened 2019-01-31
Reduce exposed use cases on Proms system

aminur80:
- separate url pattern for CIC_PROMS


rdrf/rdrf/urls.py
```diff
@@ -114,8 +114,37 @@ def handler_application_error(request):
     re_path(r'^account/two_factor/disable/?$', DisableView.as_view(), name='disable'),
 ]
 
+proms_pattern = [
```
id2359: proms_patterns

PR #766, opened 2019-02-05
Remove swipe

aminur80:
- remove easy-swipe package
- introduce react-icons (using Github Octicons)
- add GoChevronLeft & GoChevronRight icon

### DIFF
--- a/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
+++ b/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
@@ -87,8 +87,9 @@ class Question extends React.Component<QuestionInterface, object> {
 
     render() {
         let question = this.props.questions[this.props.stage];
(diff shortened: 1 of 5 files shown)


rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
```diff
@@ -87,8 +87,9 @@ class Question extends React.Component<QuestionInterface, object> {
 
     render() {
         let question = this.props.questions[this.props.stage];
-        const box_style = {width: "100px", height:"100px", backgroundColor: "black"}
-        const p_style = {color: "white", align: "center"}
+        let defaultValue = this.props.answers[question.cde] !== undefined ? this.props.answers[question.cde] : 0 ;
```
id2359: This seems to allow for the user accepting the 0 value in the widget but the event handler won't get troiggered so the answers array won't have been popuated with the 0 value. Test this edge case locally and fix if so.

PR #768, opened 2019-02-06
export/import cic clinical naming scheme

aminur80:
- additional validation for context form group  naming_scheme
- fix validation for is_default


rdrf/rdrf/models/definition/models.py
```diff
@@ -1639,14 +1639,18 @@ def naming_info(self):
...
-        defaults.exclude(pk=self.pk)
+                                                   is_default=True).exclude(pk=self.pk)
         num_defaults = defaults.count()
+
         if num_defaults > 0 and self.is_default:
             raise ValidationError("Only one Context Form Group can be the default")
         if num_defaults == 0 and not self.is_default:
             raise ValidationError("One Context Form Group must be chosen as the default")
 
+        if self.naming_scheme == "C" and self.naming_cde_to_use is None:
```
id2359: Technically this is wrong because an empty string fornaming cde would pass validation - write a mini parser method  to return a  triple of the objects needed - none otherwise 


(form_model, section_mode, cde_model)

rdrf/rdrf/models/definition/models.py
```diff
@@ -1647,10 +1647,48 @@ def clean(self):
...
 
+    def _valid_naming_cde_to_use(self, naming_cde_to_use):
+        validation_message = "Invalid naming cde: Should be form name/section code/cde code where all codes must exist"
+        if naming_cde_to_use:
+            try:
+                naming_cde_expression = naming_cde_to_use.split("/")
+                if len(naming_cde_expression) != 3:
+                    raise ValidationError(validation_message)
+                form_name, section_code, cde_code = naming_cde_expression
+            except :
```
id2359: too broad exception clause - just trap relevant error - probably a ValueError ( check in python console by attempting to destructure a 3 element tuple from None, say

aminur80: Now capturing ValueError while destructuring a tuple

id2359: Looks good now

PR #770, opened 2019-02-12
Copyright clinical

aminur80:
- adding copyright, source text at the bottom of Survey Question page
- adding a link to copyright page
- adding a new page with registry level copyright text and copyright text from individual Question


rdrf/rdrf/templates/rdrf_cdes/copyright.html
```diff
@@ -1,31 +1,19 @@
-{% load static from staticfiles %}
 
-<!DOCTYPE html>
-<html lang="en">
+{% extends "rdrf_cdes/base-1-col.html" %}
```
id2359: Good

id2359: We'll see it breaks prod build before merging.

PR #773, opened 2019-02-13
lint: first pass through

grahame:
- autopep8
- fixes a number of bugs (mostly missing imports)
- remove a number of * imports

lint will still fail but that's due to a bug I'm not sure how to fix - @id2359, maybe you could look at that once this is merged?


grahame:
Thanks @aminur80 for fixing up the tests.

PR #784, opened 2019-02-19
Proms project link

aminur80:


rdrf/rdrf/templatetags/project_title_link.py
```diff
@@ -5,9 +5,13 @@
...
 register = template.Library()
 
+_HOME_PAGE = "admin:index"
+
 
 @register.simple_tag
 def project_title_link():
-    if settings.SYSTEM_ROLE is not SystemRoles.CIC_PROMS:
+    if settings.SYSTEM_ROLE is SystemRoles.CIC_PROMS:
+        return reverse(_HOME_PAGE)
```
id2359: It's really settings.PROJECT_TITLE_LINK that should change
because now for CIC PROMS we're effectively ignoring the PROJECT_TITLE_LINK setting

Issue #869, opened 2019-07-17
Bump the django-useraudit requirement version


mouneyrac:
We need the fix done for #837

It is currently in 1.7.1 https://github.com/muccg/django-useraudit/releases/tag/1.7.1

Issue #837, opened 2019-05-24
reactivate functionality "lost"


id2359:
Updating FH to RDRF 5.0.4:

From Jing:
```
I've lost the reactivate function and cannot help Kerryn with her login. 

I get this instead:
<a href='/rdrf/useraudit/reactivate/69'>Activate</a>

Can you please assist as soon as possible?

Many thanks!

Jing
```

id2359:
This is the page
https://rdrf.ccgapps.com.au/ophg/admin/useraudit/loginattempt/
```

leetest3 | 0 | Oct. 26, 2017, 10:38 a.m. | Active
-- | -- | -- | --
testrphcurator | 1 | March 21, 2018, 3:25 p.m. | <a href='/ophg/useraudit/reactivate/65'>Activate</a>
leetest | 5 | Sept. 20, 2017, 2:57 p.m. | N/A
fshdcurator | 0 | Aug. 28, 2017, 4:29 p.m. | Active
lee2 | 4 | Aug. 18, 2017, 1:43 p.m. | N/A
fhcuraor | 2 | June 7, 2017, 10:06 a.m. | N/A
fh_curator_kn | 0 | May 31, 2017, 11:37 a.m. | Active

```

id2359:
The html appears ..

id2359:
This link is constructed within Django Useraudit

id2359:
The workaround is to reactivate from the users screen

mouneyrac:
Done - https://github.com/muccg/django-useraudit/pull/33

PR #870, opened 2019-07-17
#869 bump django-useraudit version

mouneyrac:
The django-useraudit last version includes the fix for the user reactivate feature #837


mouneyrac:
I am still creating a pull request even thought the build failed because the build failed for a different reason.

Note that I tested this branch and I confirm the activate linked is fixed (.i.e we do install the latest version of django-useraudit)

Issue #841, opened 2019-06-03
Convert JS calculation in Python


mouneyrac:
* Convert JS calculation in Python
* Django-admin command to be run periodically to update time dependant function.

mouneyrac:
The code has currently been tested through:
* unit tests
* manual testing from George
* django-admin update script: against all fields of staging comparing with the js results produced by a node container running the current JS code.

TODO:
* django-admin update script: against all fields of prod comparing with the js results produced by a node container running the current JS code.

mouneyrac:
closing - it is now on staging for testing.

mouneyrac:
List of calculations:
1- Age of consent
2- Age of assessment
3- Age of Death
4- BMI
5- LDL-cholesterol adjusted for treatment
6- Dutch Lipid Clinic Network Score (Phenotypic)
7- FH Diagnostic Category

PR #891, opened 2019-08-12
#841 move calculated_functions.py from scripts folder to forms/fields…

mouneyrac:
… folder


rdrf/rdrf/forms/dynamic/calculated_fields.py
```diff
@@ -32,7 +32,7 @@ def get_script(self):
             self.section.code + settings.FORM_SECTION_DELIMITER
         observer_code = self.cde.code
 
-        mod = __import__('rdrf.scripts.calculated_functions', fromlist=['object'])
+        mod = __import__('rdrf.forms.fields.calculated_functions', fromlist=['object'])
```
id2359: Looks good

Issue #925, opened 2019-08-23
POEM: update_calculated_fields crashes on POEM


mouneyrac:
Tested on a local machine with staging data.

root@d59e3a8077d2:~# django-admin update_calculated_fields
[INFO:2019-08-23 10:55:21,299:apps.py:14:ready] running RDRFConfig.ready ... 
[DEBUG:2019-08-23 10:55:23,917:calculated_functions.py:664:poemScore] answer: [4, 2, 4, 4, 0, 0, 0]
[DEBUG:2019-08-23 10:55:23,917:calculated_functions.py:666:poemScore] fixed_answer: [4, 2, 4, 4, 0, 0, 0]
INFO UPDATING DB: These are the new value of the form/context {'poemScore': {'old_value': '13 ( Moderate eczema )', 'new_value': '14 ( Moderate eczema )', 'section_code': 'poemSection'}} - registry: poem - patient: 667 - form: PoemScoreForm - context: 381
Traceback (most recent call last):
  File "/env/bin/django-admin", line 11, in <module>
    sys.exit(execute_from_command_line())
  File "/env/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/env/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/env/lib/python3.7/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/env/lib/python3.7/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/app/rdrf/rdrf/management/commands/update_calculated_fields.py", line 116, in handle
    registry_model)
  File "/app/rdrf/rdrf/management/commands/update_calculated_fields.py", line 175, in save_new_calculation
    skip_bad_key=True)
  File "/app/rdrf/registry/patients/models.py", line 609, in set_form_value
    raise Exception("context model should not be explicit for non-supporting registry")
Exception: context model should not be explicit for non-supporting registry

PR #927, opened 2019-08-23
#925 update_calculated_fields: context is None before saving for no c…

mouneyrac:
…ontext

POEM update_calculated_fields was crashing because context id had a value but the registry didn't
support context.


mouneyrac:
Changing the way we fix the issue - we are going to change set_form_value to support a context model when a registry has no context feature. We just add a sanity check to confirm the passed context_model is the correct one.

Issue #898, opened 2019-08-15
update_calculated_fields: remove dependency on the field having calculation


mouneyrac:
Currently we retrieve the calculated fields by their JS calculation. Instead we should retrieve them by their cde datatype: calculated

mouneyrac:
what is CDE00023 - Dutch Lipid Clinic Network Criteria Score ? It is marked a calculated field but there is no calculation. It is also in no form. Finally, it is referenced in the git repository in the file named testing_rdrf.default.json

Can you run these commands in prod to confirm that my changes in https://github.com/muccg/rdrf/commit/53eb4edaf392d7daafc565263a03a1ede11bcc08 will behave the same => .i.e. ALL calculated fields have a calculation and ONLY calculated fields have calculation:

calculated_cde_models = CommonDataElement.objects.exclude(calculation='')
calculated_cde_models2 = CommonDataElement.objects.filter(datatype='calculated')
len(calculated_cde_models) == len(calculated_cde_models2)

on staging it is not the case as CDE00023 is marked as calculated datatype but has not calculation (and it is not used anyway)

PR #928, opened 2019-08-23
#898 update_calculated_fields: retrieve calculated fields by datatype

mouneyrac:
We previously retrieved the calculated fields checking if the calculation
was empty or not.

### DIFF
--- a/rdrf/rdrf/management/commands/update_calculated_fields.py
+++ b/rdrf/rdrf/management/commands/update_calculated_fields.py
@@ -48,7 +48,7 @@ class Command(BaseCommand):
             calculated_cde_models = CommonDataElement.objects.filter(code__in=options['cde_code'])
         else:
             # Retrieve all calculated fields.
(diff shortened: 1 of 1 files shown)


mouneyrac:
what is CDE00023 - Dutch Lipid Clinic Network Criteria Score ? It is marked a calculated field but there is no calculation. It is also in no form. Finally, it is referenced in the git repository in the file named testing_rdrf.default.json

Can you run these commands in prod to confirm that my changes in 53eb4ed will behave the same => .i.e. ALL calculated fields have a calculation and ONLY calculated fields have calculation:

calculated_cde_models = CommonDataElement.objects.exclude(calculation='')
calculated_cde_models2 = CommonDataElement.objects.filter(datatype='calculated')
len(calculated_cde_models) == len(calculated_cde_models2)

on staging it is not the case as CDE00023 is marked as calculated datatype but has not calculation (and it is not used anyway)

id2359:
We will need to check all the definition files for calculated fields.

PR #942, opened 2019-08-28
#939 enforce REQUIRE_2FA with environment variable (site wide)

mouneyrac:


rdrf/registry/common/middleware.py
```diff
@@ -30,7 +31,9 @@ def process_request(self, request):
         if user is None or user.is_anonymous:
             return None
 
-        if not user.is_verified() and user.require_2_fact_auth:
+        site_requires_2fa = True if EnvConfig().get("require_2fa", False) == 1 else False
```
id2359: Why not just: site_requires_2fa = EnvConfig().get("require_2fa", False) == 1

mouneyrac: obviously :)

mouneyrac: thanks

mouneyrac: done

mouneyrac:
ready to be merged :)

PR #917, opened 2019-08-21
#911 survey: only delete surveys if successfully retrieved

mouneyrac:

PR #947, opened 2019-08-30
update calculated fields when pulling proms (to be integrated after PR #917)

mouneyrac:
To be integrated after https://github.com/muccg/rdrf/pull/917 (only the last commit is from this issue)


mouneyrac:
this PR contains old uncessary commit (I expected them to disappear once they were merged into next_release) - creating a new clean PR

PR #951, opened 2019-08-30
# 901 switch from yaml.load to yaml.safe_load (for none complex python object)

mouneyrac:
What has not been changed (because it would likely require a rewrite as it uses complex yaml file):

importer.py
def load_yaml(self, yaml_data_file):

run_yaml_cde_calculation_tests.py
def load_yaml(file_obj):

### DIFF
--- a/rdrf/rdrf/management/commands/check_structure.py
+++ b/rdrf/rdrf/management/commands/check_structure.py
@@ -75,7 +75,7 @@ class Command(BaseCommand):
 
         if os.path.exists(schema_path):
             with open(schema_path) as sf:
-                return yaml.load(sf)
+                return yaml.safe_load(sf)
 
         raise FileNotFoundError(errno.ENOENT,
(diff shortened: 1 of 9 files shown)


mouneyrac:
I am not sure these changes are really useful apart if some third party have access to the yaml files.

mouneyrac:
After thinking, because it would be hard to test each cases where we load yaml and because the way we load yaml is not accessible by third party entities if may be worth to delay the merging. It would still be good to merge but we may break the software and unfortunately the piece of codes using yaml.load() are not straight forward (.i.e. they require good knowledge and we may have more urgent tickets to deal). I would suggest we cancel this pull request and move the issue back to triage required/backlog for the moment.

mouneyrac:
I am going to close this pull request. Each convertion have not be 100% (.i.e. we can not detect errors but we are not sure that the changes do not break anything as we don't have a full understanding of the impact of each changes).

Issue #960, opened 2019-09-04
Calculated fields: Traceback in registry.log on staging for rdrf-ophg


id2359:
```
[ERROR:2019-09-04 10:25:35,861:log.py:228:log_response] Internal Server Error: /ophg/api/v1/calculatedcdes/
Traceback (most recent call last):
  File "/env/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/env/lib/python3.7/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/env/lib/python3.7/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/env/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/env/lib/python3.7/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/env/lib/python3.7/site-packages/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/env/lib/python3.7/site-packages/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/env/lib/python3.7/site-packages/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/env/lib/python3.7/site-packages/rdrf/services/rest/views/api_views.py", line 311, in post
    return Response(func(patient_values, form_values))
  File "/env/lib/python3.7/site-packages/rdrf/forms/fields/calculated_functions.py", line 525, in fhAgeAtConsent
    consentDate = datetime.strptime(context["FHconsentDate"], '%Y-%m-%d')
  File "/usr/local/lib/python3.7/_strptime.py", line 577, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "/usr/local/lib/python3.7/_strptime.py", line 359, in _strptime
    (data_string, format))
ValueError: time data '01-03-19' does not match format '%Y-%m-%d'
[ERROR:2019-09-04 10:25:35,873:log.py:228:log_response] Internal Server Error: /ophg/api/v1/calculatedcdes/
Traceback (most recent call last):
[...]

PR #972, opened 2019-09-06
#960 calculated field ws: return 400 bad request for bad dates

mouneyrac:


mouneyrac:
need a little bit more work...

Issue #959, opened 2019-09-04
Demo site: exception when logging with admin


mouneyrac:
NoReverseMatch at /patientslisting
Reverse for 'parent_page' not found. 'parent_page' is not a valid view function or pattern name.
Request Method:	GET
Request URL:	https://rdrf.ccgapps.com.au/demo/patientslisting
Django Version:	2.1.9
Exception Type:	NoReverseMatch
Exception Value:	
Reverse for 'parent_page' not found. 'parent_page' is not a valid view function or pattern name.
Exception Location:	/env/lib/python3.7/site-packages/django/urls/resolvers.py in _reverse_with_prefix, line 622
Python Executable:	/env/bin/uwsgi
Python Version:	3.7.4
Python Path:	
['.',
 '/app',
 '/usr/local/lib/python37.zip',
 '/usr/local/lib/python3.7',
 '/usr/local/lib/python3.7/lib-dynload',
 '/env/lib/python3.7/site-packages']
Server time:	Wed, 4 Sep 2019 10:00:04 +0800
Error during template rendering
In template /env/lib/python3.7/site-packages/rdrf/templates/rdrf_cdes/base.html, error at line 123

Reverse for 'parent_page' not found. 'parent_page' is not a valid view function or pattern name.
113	
114	                
115	                <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
116	                    <ul class="nav navbar-nav navbar-right">
117	                        {% if request.user.is_authenticated %}
118	                        {% block parent_menu %}
119	                        {% endblock %}
120	
121	                        {% if request.user.registry_code %}
122	                        {% if request.user.is_parent %}
123	                        <li><a href="{% url 'parent_page' request.user.registry_code %}"><span class="glyphicon glyphicon-home"></span>
124	                                {% trans "Home Page" %}</a></li>
125	                        {% elif request.user.is_patient %}
126	                        <li><a href="{% url 'patient_page' request.user.registry_code %}"><span class="glyphicon glyphicon-home"></span>
127	                                {% trans "Home Page" %}</a></li>
128	                        {% endif %}
129	                        {% endif %}
130	
131	                        {% block actions %}
132	                        {% endblock %}
133	
Traceback Switch to copy-and-paste view
/env/lib/python3.7/site-packages/django/core/handlers/exception.py in inner
            response = get_response(request) ...
▶ Local vars
/env/lib/python3.7/site-packages/django/core/handlers/base.py in _get_response
[...]

PR #974, opened 2019-09-09
#959 ignore is_parent for anything else than angelman

mouneyrac:
This fix is a little between two chairs:
- it works by checking if the parent_view module exists which I think is an ok check.
- however it hardcodes angleman module location because the parent_view module is in the angelman part, not in rdrf.

PS: tested on rdrf and Angelman. On Angelman the parent login unit test is executed and the is_parent() function is executed too.

### DIFF
--- a/rdrf/rdrf/testing/unit/tests.py
+++ b/rdrf/rdrf/testing/unit/tests.py
@@ -1573,10 +1573,17 @@ class RemindersTestCase(TestCase):
         assert result == "testuser\n", "Expected testuser instead got [%s]" % result
 
         # parents are detected
-        self._setup_user("testuser", Time.LONG_AGO, group="parents")
-        result = self._run_command(registry_code="foobar", days=365)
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/testing/unit/tests.py
```diff
@@ -1573,10 +1573,17 @@ class Time:
...
 
         # parents are detected
-        self._setup_user("testuser", Time.LONG_AGO, group="parents")
-        result = self._run_command(registry_code="foobar", days=365)
-        assert result == "testuser\n", "Expected testuser instead got [%s]" % result
-
+        parent_feature = True
+        # Check parent feature is enabled.
+        try:
+            __import__('angelman.parent_view')
```
id2359: As a stopgap solution this is ok, but we should try to eliminate having to do this in future by ultimately merging angelman into base rdrf

PR #989, opened 2019-09-20
#981 bump django version for security purpose

mouneyrac:


mouneyrac:
I did a quick manual smoke test and I didn't detect any errors (calculate fields, proms, administration, save oprations...)

PR #998, opened 2019-09-27
Cleanup logging

mouneyrac:


rdrf/rdrf/workflows/verification.py
```diff
@@ -146,10 +146,10 @@ def has_annotation(self, user, registry_model, patient_model, context_model):
         the cde has changed?
         """
         def carp(msg):
-            logger.debug("Annotations Patient %s Context %s CDE %s: %s" % (patient_model,
-                                                                           context_model.id,
-                                                                           self.cde_model.code,
-                                                                           msg))
+            logger.info("Annotations Patient %s Context %s CDE %s: %s" % (patient_model,
```
id2359: better to use patient_model.id

rdrf/rdrf/auth/forms.py
```diff
@@ -139,8 +139,8 @@ def save(self, commit=True):
                     self.user.is_active = True
             else:
                 logger.warning(
-                    'User "%s" resetted their password but their account is inactive '
-                    'and settings.ACCOUNT_SELF_UNLOCK_ENABLED is NOT set.', self.user)
+                    'User id "%s" resetted their password but their account is inactive '
+                    'and settings.ACCOUNT_SELF_UNLOCK_ENABLED is NOT set.', self.user.id)
```
id2359: For Midland they are actually interested in knowing who did what so I think recording the username is ok - it will be difficult for them to find the corresponding user from the id. For patients , it is different, ids are good in that case.

rdrf/rdrf/helpers/utils.py
```diff
@@ -787,7 +787,7 @@ def is_authorised(user, patient_model):
     if common and not user.is_parent:
         return True
 
-    logger.warning("user %s is not authorised for patient %s" % (user, patient_model.pk))
+    logger.warning("user id %s is not authorised for patient %s" % (user.id, patient_model.pk))
```
id2359: same as above

rdrf/rdrf/workflows/verification.py
mouneyrac: hum not sure how you ended up seen this line... I must have created the pull request too early and pushed a change later as it has been fixed https://github.com/muccg/rdrf/pull/998/files#diff-a751215b7458437a4864aa8d56b3917fR149 -

rdrf/rdrf/auth/forms.py
mouneyrac: no problem I'll make the change.

rdrf/rdrf/helpers/utils.py
mouneyrac: no problem I'll make the change.

rdrf/rdrf/helpers/utils.py
```diff
@@ -788,7 +788,7 @@ def is_authorised(user, patient_model):
     if common and not user.is_parent:
         return True
 
-    logger.warning("user id %s is not authorised for patient %s" % (user.id, patient_model.pk))
+    logger.warning("user %s is not authorised for patient %s" % (user.username, patient_model.pk))
```
id2359: Thanks

PR #999, opened 2019-09-30
Support for syslog

mouneyrac:


rdrf/rdrf/settings.py
```diff
@@ -337,6 +337,22 @@
...
 
+# Log handlers
+DEFAULT_LOG_HANDLER = ['console', 'file']
+MAILADMIN_LOG_HANDLER = ['mail_admins']
+COMMAND_LOG_HANDLER = ['shell', 'admin_command_file']
+IMPORT_LOG_HANDLER = ['console_simple']
+
+# Add the syslog handler when syslog is enabled.
+SYSLOG_ENABLED = env.get("syslog_enabled", False)
+SYSLOG_ADDRESS = env.get("syslog_address", "172.21.0.1")
```
id2359: Don't think we should have a default ip here. 

>

mouneyrac: this is the default docker host ip, I would argue that it is actually a good idea to leave it here for DX as most dev won't have the pain to find out for themself when they install rsyslog in their host.

PR #1000, opened 2019-10-01
ldap support

mouneyrac:


rdrf/rdrf/auth/ldap_backend.py
```diff
@@ -13,25 +14,35 @@ class RDRFLDAPBackend(LDAPBackend):
...
+        # We need to update the user information whatever it existed or not in RDRF database.
+
+        from rdrf.models.definition.models import Registry
+        from registry.groups.models import WorkingGroup
+        registry_model = Registry.objects.get(code=settings.RDRF_AUTH_LDAP_REGISTRY_MODEL)
+        wg = WorkingGroup.objects.get(registry=registry_model,
+                                      name=settings.RDRF_AUTH_LDAP_WORKING_GROUP)
+
+        # Check if user must be staff anyway
+        if settings.RDRF_AUTH_LDAP_USERS_MUST_BE_STAFF:
```
id2359: All users must be staff so I think  the setting is redundant - being staff is forced.

rdrf/rdrf/settings.py
```diff
@@ -202,46 +203,64 @@
...
+AUTH_LDAP_BIND_DN = env.get("auth_ldap_bind_dn", "cn=admin,dc=example,dc=com")
+AUTH_LDAP_BIND_PASSWORD = env.get("auth_ldap_bind_password", "admin")
+# enviroment variables prefixed RDRF_ to not conflict with the variables expected by django auth ldap plugin
+RDRF_AUTH_LDAP_BIND_DC = env.get("rdrf_auth_ldap_bind_dc", "dc=example,dc=com")
+RDRF_AUTH_LDAP_BIND_GROUP = env.get("rdrf_auth_ldap_bind_group", "ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_FIRST_NAME_ATTR = env.get("rdrf_auth_ldap_first_name_attr", "sn")
+RDRF_AUTH_LDAP_LAST_NAME_ATTR = env.get("rdrf_auth_ldap_last_name_attr", "sn")
+RDRF_AUTH_LDAP_MAIL_ATTR = env.get("rdrf_auth_ldap_mail_attr", "mail")
+RDRF_AUTH_LDAP_IS_ACTIVE_GROUP = env.get("rdrf_auth_ldap_is_active_group", "cn=active,ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_IS_STAFF_GROUP1 = env.get("rdrf_auth_ldap_is_staff_group1", "cn=staff,ou=groups,dc=example,dc=com")
```
id2359: What are group 1 and group2 ?

rdrf/rdrf/settings.py
```diff
@@ -202,46 +203,64 @@
...
 # apps use modelbackend by default, but can be overridden here
 # see: https://docs.djangoproject.com/en/dev/ref/settings/#authentication-backends
 
 AUTHENTICATION_BACKENDS = [
-    'rdrf.auth.ldap_backend.RDRFLDAPBackend',
     'useraudit.password_expiry.AccountExpiryBackend',
     'django.contrib.auth.backends.ModelBackend',
     'useraudit.backend.AuthFailedLoggerBackend',
 ]
+if LDAP_ENABLED:
```
id2359: good

rdrf/rdrf/settings.py
```diff
@@ -202,46 +203,64 @@
...
+RDRF_AUTH_LDAP_BIND_GROUP = env.get("rdrf_auth_ldap_bind_group", "ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_FIRST_NAME_ATTR = env.get("rdrf_auth_ldap_first_name_attr", "sn")
+RDRF_AUTH_LDAP_LAST_NAME_ATTR = env.get("rdrf_auth_ldap_last_name_attr", "sn")
+RDRF_AUTH_LDAP_MAIL_ATTR = env.get("rdrf_auth_ldap_mail_attr", "mail")
+RDRF_AUTH_LDAP_IS_ACTIVE_GROUP = env.get("rdrf_auth_ldap_is_active_group", "cn=active,ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_IS_STAFF_GROUP1 = env.get("rdrf_auth_ldap_is_staff_group1", "cn=staff,ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_IS_STAFF_GROUP2 = env.get("rdrf_auth_ldap_is_staff_group2", "cn=admin,ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_IS_SUPERUSER_GROUP = env.get("rdrf_auth_ldap_is_superuser_group", "cn=superuser,ou=groups,dc=example,dc=com")
+RDRF_AUTH_LDAP_POSTFIXGROUP_ATTR = env.get("rdrf_auth_ldap_postfixgroup_attr", "cn")
+RDRF_AUTH_LDAP_REGISTRY_MODEL = env.get("rdrf_auth_ldap_registry_model", "ICHOMCRC")
```
id2359: RDRF_AUTH_LDAP_REGISTRY_CODE
rdrf_auth_ldap_registry_code

rdrf/rdrf/auth/ldap_backend.py
mouneyrac: Good point. In this case I will remove the staff group binding too. Actually it is better this way, if ever we need these settings we can add them back. But it is unlikely anyway.

rdrf/rdrf/settings.py
mouneyrac: I will remove staff groups.

rdrf/rdrf/settings.py
mouneyrac: thanks, I'll change it.

Issue #905, opened 2019-08-19
check if we log raised exceptions by the django-admin commands


mouneyrac:
That would be useful to quickly check which exception has been raised. Maybe it already exist?

mouneyrac:
this is how to log all exceptions from django-admin (likely what we should do):
https://mpcabd.xyz/make-django-management-commands-send-errors-email-admins/

mouneyrac:
in fact django-admin does not run manage.py so we needed another way to log exceptions from django-admin. I created a decorator and added it to update_calculated_fields command line.

PR #1003, opened 2019-10-01
#905 catch and log exceptions from update_calculated_fields

mouneyrac:
Otherwise exceptions would not be logged when running update_calculated_fields with django-admin command line.


rdrf/rdrf/management/commands/update_calculated_fields.py
```diff
@@ -35,6 +36,7 @@ def add_arguments(self, parser):
         # Test command line example
         # django-admin update_calculated_fields --patient_id=2 --registry_code=fh --form_name=ClinicalData --section_code=SEC0007 --context_id=2 --cde_code=CDEfhDutchLipidClinicNetwork
 
+    @catch_and_log_exceptions
```
id2359: nice

Issue #1007, opened 2019-10-03
Convert env defaults into their own variables


mouneyrac:
For example 

AUTH_LDAP_SERVER_URI = env.get("auth_ldap_server_uri", "ldap://ldap")

=> 

LDAP_URL = "ldap://ldap"
AUTH_LDAP_SERVER_URI = env.get("auth_ldap_server_uri", LDAP_URL)

PR #1009, opened 2019-10-03
#1007 Convert env defaults into their own variables

mouneyrac:


mouneyrac:
some tweaks to be done...

Issue #976, opened 2019-09-13
Harden reporting interface


id2359:
all queries must start with:  "SELECT p.id"     ( this is required to make the report work anyway)
"UNION " must not be in the clause
"UPDATE " must not be the clause
"DROP " must not be in the clause

case insensitive

PR #1023, opened 2019-10-11
#976 explorer: harden reporting interface

mouneyrac:
all queries must start with: "SELECT p.id" ( this is required to make the report work anyway)
"UNION " must not be in the clause
"UPDATE " must not be the clause
"DROP " must not be in the clause


mouneyrac:
duplicate of https://github.com/muccg/rdrf/pull/1070 (.i.e. I made the requested change there)

Issue #1031, opened 2019-10-16
Baseline proms blank screen after a certain answer sequence


id2359:
answers to first five questions

1. tertiary
2. married/partnered
3. caucasian
4. employed
5 manager/professional

then blank sscreen 

control shift j shows error: t.spec is null

PR #1032, opened 2019-10-16
#1031 proms: support for input fields

mouneyrac:


mouneyrac:
sorry for this part: https://github.com/muccg/rdrf/pull/1032/commits/a6ebaf0cef11f69e4eec54f200f9a1d13bc191ba#diff-f85efff41617d9bf196aed33a657b923R66-R101

It is just pure linting that happens when I saved.

Issue #1050, opened 2019-10-21
[android] proms: can not scroll down the vertical slider (.i.e. can never finish the proms)


mouneyrac:
Desktop
1- Login as admin
2- Create a QR proms for a user

Android - LG V30+
Chrome
3- Use Google lens as QR code
4- Go to Your Health Today question

=> you can not scroll down, forbidding you to continue answering the proms.

PR #1055, opened 2019-10-22
#1050 fix proms layout

mouneyrac:
- add back vertical scrolling
- remove unused 100% div at the bottom
- fix button positioning + fix button size consistency

Still not perfect but way better on mobile.


mouneyrac:
After a second thought - the rest of the file use inline style so I think it is fine.

Issue #1072, opened 2019-11-01
Template does not exists (empty template name)


mouneyrac:
Internal Server Error: /rdrf/ang/register/

TemplateDoesNotExist at /ang/register/
.html

Request Method: POST
Request URL: https://angelmanregistry.info/rdrf/ang/register/
Django Version: 2.1.7
Python Executable: /env/bin/uwsgi
Python Version: 3.7.4
Python Path: ['.', '/app', '/usr/local/lib/python37.zip', '/usr/local/lib/python3.7', '/usr/local/lib/python3.7/lib-dynload', '/env/lib/python3.7/site-packages', '/env/src/django-rdrf/rdrf', '/app/angelman']
Server time: Thu, 31 Oct 2019 17:48:00 +0800
Installed Applications:
['angelman',
 'django.contrib.contenttypes',
 'django.contrib.auth',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.staticfiles',
 'django.contrib.messages',
 'django_extensions',
 'django.contrib.admin',
 'messages_ui',
 'ajax_select',
 'explorer',
 'useraudit',
 'templatetag_handlebars',
 'iprestrict',
 'rest_framework',
 'anymail',
 'rdrf',
 'registry.groups',
 'registry.patients',
 'registry.common',
 'registry.genetic',
 'registration',
 'reversion',
 'storages',
 'django_otp',
 'django_otp.plugins.otp_static',
 'django_otp.plugins.otp_totp',
 'two_factor',
 'django_user_agents',
 'formtools']
Installed Middleware:
('useraudit.middleware.RequestToThreadLocalMiddleware',
 'django.middleware.common.CommonMiddleware',
 'iprestrict.middleware.IPRestrictMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django_otp.middleware.OTPMiddleware',
 'registry.common.middleware.EnforceTwoFactorAuthMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django_user_agents.middleware.UserAgentMiddleware')
Template loader postmortem
Django tried loading these templates, in this order:

Using engine django:
[...]

PR #1073, opened 2019-11-01
#1072 registration: backend form validations do not crash anymore

mouneyrac:


mouneyrac:
As I am not expert in Django, I would understand there may be a better solution.

Basically I switched the "return self.form_invalid" to "return render", which is what we do in other places in RDRF.
I have not seen this "return self.form_invalid" (on the base class as in this code) neither in our code or in the documentation - we do have one "return self.form_invalid" in our code but it is intercepted by our own code and transform in render. So I changed for "return render".

Once I did it I found out the problem came from the missing information about the registry_code.

mouneyrac:
Tested on Angelman.

rdrf/rdrf/views/registration_rdrf.py
```diff
@@ -45,7 +45,9 @@ def post(self, request, *args, **kwargs):
         if form.is_valid():
             return self.form_valid(form)
         else:
-            return self.form_invalid(form)
+            context = self.get_context_data(form=form)
```
id2359: Add logging form validation errors here ( for error in form.errors )

mouneyrac: done

mouneyrac: example of log:
runserver_1    | [WARNING:2019-11-04 15:55:46,367:registration_rdrf.py:56:post] Backend validation of the registration has failed: {'username': [{'message': 'Custom user with this Username already exists.', 'code': 'unique'}]}

PR #1105, opened 2019-11-27
Rdrf 1094 v2

mouneyrac:


mouneyrac:
needs to clean up this PR (plenty of commit that should not be here)

Issue #1107, opened 2019-11-28
Rare Lipid calcs need converting to python


id2359:
[<CommonDataElement: CDE CDEBMI:BMI>, <CommonDataElement: CDE FHDeathAge:Age at death>, <CommonDataElement: CDE RLDRDeathAge:Age at death>]
>>>

jithendralal:
The function _is_calculated_field made to look in cde.datatype instead of cde.calculation.

PR #1123, opened 2019-12-11
rdrf #1107 RLDRDeathAge calculation added

jithendralal:
- RLDRDeathAge calculation added
- funcrion _is_calculated_field made to look in cde.datatype instead of cde.calculation

### DIFF
--- a/rdrf/rdrf/forms/dynamic/field_lookup.py
+++ b/rdrf/rdrf/forms/dynamic/field_lookup.py
@@ -171,7 +171,7 @@ class FieldFactory(object):
         return False
 
     def _is_calculated_field(self):
-        return bool(self.cde.calculation)
+        return self.cde.datatype == 'calculated'
 
     def _is_complex(self):
         return self.complex_field_factory._is_complex()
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/forms/dynamic/field_lookup.py
```diff
@@ -171,7 +171,7 @@ def _has_other_please_specify(self):
         return False
 
     def _is_calculated_field(self):
-        return bool(self.cde.calculation)
+        return self.cde.datatype == 'calculated'
```
id2359: This needs to be checked against existing registries

rdrf/rdrf/forms/fields/calculated_functions.py
```diff
@@ -513,6 +513,30 @@ def FHDeathAge_inputs():
 
 ################ END OF FHDeathAge ################################
 
+################ BEGINNING OF RLDRDeathAge ################################
+
+def RLDRDeathAge(patient, context):
```
id2359: This code is dead , in fact - so remove ads FHDeathAge is the cde being used ...

Issue #1067, opened 2019-10-31
User should not be able to access Django registration page /account/register


id2359:
in core rdrf, user should not be able to access this page (remove it):
![normal rdrf - account register url](https://user-images.githubusercontent.com/354908/68096851-85aeae80-feee-11e9-93be-46cf53947dab.png)

(description updated to reflect the new topic of this issue - Jerome)

id2359:
TemplateDoesNotExist at /ang/register/
.html
```

**Comment by id2359** [comment]
```
Traceback:

File "/env/lib/python3.7/site-packages/django/template/backends/django.py" in render
  61.             return self.template.render(context)

File "/env/lib/python3.7/site-packages/django/template/base.py" in render
  171.                     return self._render(context)

File "/env/lib/python3.7/site-packages/django/template/base.py" in _render
  163.         return self.nodelist.render(context)

File "/env/lib/python3.7/site-packages/django/template/base.py" in render
  937.                 bit = node.render_annotated(context)

File "/env/lib/python3.7/site-packages/django/template/base.py" in render_annotated
  904.             return self.render(context)

File "/env/lib/python3.7/site-packages/django/template/defaulttags.py" in render
  512.             return self.nodelist.render(context)

File "/env/lib/python3.7/site-packages/django/template/base.py" in render
  937.                 bit = node.render_annotated(context)

File "/env/lib/python3.7/site-packages/django/template/base.py" in render_annotated
  904.             return self.render(context)

File "/env/lib/python3.7/site-packages/django/template/loader_tags.py" in render
  176.                 template = context.template.engine.get_template(template_name)

File "/env/lib/python3.7/site-packages/django/template/engine.py" in get_template
  144.         template, origin = self.find_template(template_name)

File "/env/lib/python3.7/site-packages/django/template/engine.py" in find_template
  130.         raise TemplateDoesNotExist(name, tried=tried)

The above exception (.html) was the direct cause of the following exception:

File "/env/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/env/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  156.                 response = self.process_exception_by_middleware(e, request)

File "/env/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  154.                 response = response.render()

File "/env/lib/python3.7/site-packages/django/template/response.py" in render
  106.             self.content = self.rendered_content

[...]

PR #1128, opened 2019-12-17
rdrf #1067 registration page made hidden

jithendralal:
- The /accounts/register url made inaccessible for visitor. 
- The register url will still work if used like /[registry-code]/register (if a registry with code registry-code exists)

### DIFF
--- a/rdrf/rdrf/views/registration_rdrf.py
+++ b/rdrf/rdrf/views/registration_rdrf.py
@@ -1,8 +1,10 @@
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.db import transaction
-from django.shortcuts import render
+from django.http import Http404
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/views/registration_rdrf.py
```diff
@@ -15,6 +17,11 @@ class RdrfRegistrationView(RegistrationView):
     registry_code = None
 
     def get(self, request, *args, **kwargs):
+        try:
+            get_object_or_404(Registry, code=kwargs['registry_code'])
```
id2359: I think we should test for the registry having the feature 'registration' too:

reg = get_object_or_404(Registry, code=kwargs['registry_code'])
if not reg.has_feature('registration'):
      raise Http404

PR #1205, opened 2020-02-05
Cic eq5d

jithendralal:
EQ5D related changes were made
- Health today slider labels modified
- The alignment of instructions adjusted
- The best and worst words in the instruction underlined for Health Today slider

### DIFF
--- a/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
+++ b/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
@@ -10,7 +10,6 @@ import { QuestionInterface } from './interfaces';
 import * as actions from '../reducers';
 
 
-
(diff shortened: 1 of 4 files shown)


rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
```diff
@@ -196,19 +223,32 @@ class Question extends React.Component<QuestionInterface, object> {
             return this.renderMultiSelect(question);
         }
 
+        let transformedInstruction: string[];
```
id2359: could try "disjunctive type" expression: string[] | string

jithendralal: The disjunctive type annotation has been used and it worked.

Issue #1245, opened 2020-03-06
Selective italicising of survey question instructions


jithendralal:
The modification requested by the EQ5D team:

Show the following in italics:
(e.g. work, study, housework, family or leisure activities)

PR #1248, opened 2020-03-09
rdrf #1245 

jithendralal:
Survey question instructions made italic for selected question

### DIFF
--- a/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
+++ b/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
@@ -39,6 +39,11 @@ class Question extends React.Component<QuestionInterface, object> {
             mainArray = mainString;
         }
         switch(transformation) {
+            case 'italic':
+                for (const substring of mainArray) {
+                    result.push(' ', <i>{substring}</i>);
+                }
+                return result;
             case 'underline':
                 for (const substring of mainArray) {
                     if (words.includes(substring)) {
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
```diff
@@ -39,6 +39,11 @@ class Question extends React.Component<QuestionInterface, object> {
             mainArray = mainString;
         }
         switch(transformation) {
+            case 'italic':
+                for (const substring of mainArray) {
```
id2359: If we're pipelining and mapping looking up "fluent" interfaces and functional idioms like map, filter - maybe if we refactor later if this becomes more complex

rdrf/rdrf/settings.py
```diff
@@ -315,11 +315,23 @@
...
 DEFAULT_FROM_EMAIL = env.get('default_from_email', 'No Reply <no-reply@mg.ccgapps.com.au>')
 SERVER_EMAIL = env.get('server_email', DEFAULT_FROM_EMAIL)
-EMAIL_BACKEND = 'anymail.backends.mailgun.EmailBackend'
-
-ANYMAIL = {
-    'MAILGUN_API_KEY': env.get('DJANGO_MAILGUN_API_KEY', ''),
-}
+EMAIL_BACKEND = env.get('EMAIL_BACKEND', 'anymail.backends.mailgun.EmailBackend')
+
+if EMAIL_BACKEND == 'smtp':
```
id2359: Indicate somewhere in the ticket for this that this is referring to another ticker

PR #1274, opened 2020-03-26
Cde usage 1271

jithendralal:
CDE usage displayed.


rdrf/rdrf/models/definition/models.py
```diff
@@ -752,6 +755,11 @@ def _get_typed_value(self, value):
             return float(value)
         return value
 
+    def get_usage(self):
+        sections = Section.objects.filter(elements__contains=self.code)
```
id2359: wrong for same reason as before..

Issue #1273, opened 2020-03-26
Add new feature for testing CIC CRC


gmaloneccg:
Adding a new feature for testing basic functions for CIC CRC.
- Check PROMS GUI working properly;
  - [x] Check list of PROMS request options is correct
  - [x] Check list of transmission options is correct
  - [x] ~Make a request~
  - [x] ~Check if the request succeeded, e.g. check length of PROMS requests list~
(latter two completed and tested but removed due to potential clutter from multiple PROMS requests)

gmaloneccg:
Step to check PROMS request options combined -- step checks both the survey type options and transmission method options.

PR #1275, opened 2020-03-27
rdrf #1273 Added CIC CRC feature

gmaloneccg:
New feature to test if all patient data pages are accessible to a clinical staff user, and to check if the PROMS options are behaving correctly. New step to check the PROMS options, new function to support step.

### DIFF
--- a/rdrf/rdrf/testing/behaviour/features/steps.py
+++ b/rdrf/rdrf/testing/behaviour/features/steps.py
@@ -816,15 +816,15 @@ def find(xp):
     return element
 
 
+def scroll_to_centre(xp):
+    y = find(xp).location["y"]
+    off = world.browser.get_window_size()["height"]
+    move = y - (1 / 2) * off
+    world.browser.execute_script("scrollTo(0, %s)" % move)
+
+
 @step('I add patient name "(.*)" sex "(.*)" birthdate "(.*)"')
 def add_new_patient(step, name, sex, birthdate):
-
-    def scroll_to_centre(xp):
-        y = find(xp).location["y"]
-        off = world.browser.get_window_size()["height"]
-        move = y - (1 / 2) * off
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/testing/behaviour/features/steps.py
```diff
@@ -848,3 +848,49 @@ def return_to_patientlisting(step):
...
+
+
+def find_multiple(xp):
+    """
+    Helper function to find multiple items with the same xpath.
+    Initially created so that a list of elements could be iterated over,
+    and a particular attribute extracted for each item.
+    """
+    element_list = world.browser.find_elements_by_xpath(xp)
+    return element_list
```
id2359: good - clean small functions

rdrf/rdrf/testing/behaviour/features/steps.py
```diff
@@ -848,3 +848,49 @@ def return_to_patientlisting(step):
...
+    and communication methods are correct.
+    """
+    # First check if on the PROMS page;
+    assert world.browser.current_url.endswith("/clinicalproms"),\
+        "Not on PROMS request page!"
+    # Create empty objects;
+    request_type_list = []
+    comms_type_list = []
+    # Press the add button and check if fades in;
+    find("//a[contains(@class, 'btn btn-info')]").click()
```
id2359: One thought: it might be worth naming the xpath expressions with what they're meant to find:

e.g. ADD_BUTTON = "//a[contains(@class, 'btn btn-info')]"

then find(ADD_BUTTON).click() tells us what you're trying to do

One trick I use sometimes is to emulate "enums" by grouping names as static members of a class like this:

class Colours:
    RED = 1
    BLUE = 2

so then you can say if house_colour == Colours.RED etc

In same way portions of the page could be grouped with named xpath expressions maybe

Issue #1281, opened 2020-04-01
ContextFormGroupItemAdmin 


jithendralal:
An admin user can add multiple ContextFormGroupItemAdmin.
In other words, ContextFormGroup allows admin user to add multiple forms to it.

But, this being not permitted, later causes assertion error, when a staff user logs in to the admin site and clicks a patient from the PatientList.

AssertionError: Context Form group must only contain one form

Suggested change:
Limit the number of extra forms to 0 in the StackedInline; ContextFormGroupItemAdmin (thus allowing only 1 form to be added)

jithendralal:
File "/app/rdrf/rdrf/views/patient_view.py", line 627, in get
"context_launcher": context_launcher.html,
File "/app/rdrf/rdrf/forms/components.py", line 57, in html
return self._fill_template()
File "/app/rdrf/rdrf/forms/components.py", line 64, in _fill_template
data = self._get_template_data()
File "/app/rdrf/rdrf/forms/components.py", line 131, in _get_template_data
"multiple_contexts": self._get_multiple_contexts(),
File "/app/rdrf/rdrf/forms/components.py", line 237, in _get_multiple_contexts
form.existing_links = self._get_existing_links(context_form_group)
File "/app/rdrf/rdrf/forms/components.py", line 253, in _get_existing_links
for url, text, locking in self.patient_model.get_forms_by_group(context_form_group):
File "/app/rdrf/registry/patients/models.py", line 1028, in get_forms_by_group
assert context_form_group.supports_direct_linking, "Context Form group must only contain one form"
AssertionError: Context Form group must only contain one form

id2359:
There are two considerations:  Fixed groups do allow more than one form
The direct linking idea is that if a group contains one form, a link to that form can be presented 

(  link to   [f] means link to f  

but if the group looks like

[f g] we can't present a single link to the user

jithendralal:
max_num is unlimited if context_type is Fixed and if Multiple it is limited to 1
extra is 3 if context_type is Fixed and if Multiple it is set to 0

jithendralal:
To test this:

1. Select Multiple in context type and save.
Then the items will be limited to one.
No extra forms are shown.

2. Select Fixed in context type and save.
Then there will be 3 extra forms in the items (for adding new items)
There is no limit to number of items.

gmaloneccg:
[...]

PR #1312, opened 2020-04-28
rdrf #1281 ContextFormGroupItemAdmin

jithendralal:
Limited the ContextFormGroupItems to one.

### DIFF
--- a/rdrf/rdrf/admin.py
+++ b/rdrf/rdrf/admin.py
@@ -407,6 +407,8 @@ class SurveyAssignmentAdmin(admin.ModelAdmin):
 
 class ContextFormGroupItemAdmin(admin.StackedInline):
     model = ContextFormGroupItem
+    extra = 0
+    max_num = 1
 
 
 class ContextFormGroupAdmin(admin.ModelAdmin):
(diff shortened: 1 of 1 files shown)


jithendralal:
> I don't think this will work
> e.g. see
> https://rdrf.ccgapps.com.au/ciclung/admin/rdrf/contextformgroup/1/change/
> 
> We only want to limit the items for non-fixed context groups

The changes made:
max_num is made unlimited if context_type is Fixed and if Multiple it is limited to 1
extra is 3 if context_type is Fixed and if Multiple it is limited to 0

rdrf/rdrf/admin.py
```diff
@@ -407,8 +407,18 @@ class SurveyAssignmentAdmin(admin.ModelAdmin):
...
     model = ContextFormGroupItem
-    extra = 0
-    max_num = 1
+
+    def get_max_num(self, request, obj=None, **kwargs):
+        max_num = None
+        if obj and obj.context_type == "M":
+            max_num = 1
+        return max_num
+
```
id2359, line 416: Good one

Issue #1323, opened 2020-05-19
Compare the registry definitions for differences in CDEs and Survey Questions


jithendralal:
**CIC CRC Sites Staging** vs **CIC CRC Midland Sites**
```
files: ciccrcsitestaging.yaml ciccrcmidlandsitesystem.yaml
registry definition: ICHOMCRC
versions: 1.20 1.16

Common Data Elements::
BCCAweight,abnormality_condition: ,abnormality_condition: None
BCCAweight,datatype: float,datatype: string
Country, ,Not found
EQ_Health_Rate,abnormality_condition: ,abnormality_condition: None
EQ_Health_Rate,instructions: We would like to know how good or bad your health is TODAY. This scale is numbered from 0 to 100. 100 means the best health you can imagine. 0 means the worst health you can imagine. Please tap on the scale to indicate how your health is TODAY.,instructions: We would like to know how good or bad your health is TODAY.
This scale is numbered from 0 to 100.
100 means the best health you can imagine
0 means the worst health you can imagine
Please tap on the scale to indicate how your health is TODAY.
Weightloss_amount,datatype: float,datatype: integer
CRCCompTreatment,widget_name: ,widget_name: RadioSelect
VERGE,abnormality_condition: ,abnormality_condition: None
VERGE,datatype: float,datatype: String
WEIGHTCLIN,abnormality_condition: ,abnormality_condition: None
WEIGHTCLIN,datatype: float,datatype: integer
EQ_Selfcare,abnormality_condition: ,abnormality_condition: None
RADIOTXTYPECRC,abnormality_condition: ,abnormality_condition: None
RADIOTXTYPECRC,allow_multiple: False,allow_multiple: True

Survey Questions::

In Survey: BaselinePROMS
EQ_Pain_Discomfort,instruction: Please tap the ONE box that best describes your health TODAY.,instruction: Please tick the ONE box that best describes your health TODAY
EQ_Mobility,instruction: Please tap the ONE box that best describes your health TODAY.,instruction: Please tick the ONE box that best describes your health TODAY
EQ_Anxiety_Depress,instruction: Please tap the ONE box that best describes your health TODAY.,instruction: Please tick the ONE box that best describes your health TODAY
EQ_Health_Rate,instruction: ,instruction: We would like to know how good or bad your health is TODAY. This scale is numbered from 0 to 100 (100 means the best health you can imagine 0 means the worst health you can imagine)
[...]

PR #1325, opened 2020-05-19
Script added to compare yamls

jithendralal:
The script compares CDEs and Survery Questions in the two registry definition yaml and outputs the findings to a csv file.

### DIFF
--- a/scripts/yamlcompare.py
+++ b/scripts/yamlcompare.py
@@ -1,0 +1,74 @@
+import yaml
+import sys
+import os
+
+"""
+Looks for differences in 
+    Common Data Elements and 
+    Survey Questions
+in two registry definition yaml files.
+"""
+
+
+def get_dict_of_items(yaml_data, list_to_compare, identifier):
+    return {item[identifier]: [attr + ": " + str(item[attr]) for attr in item] for item in yaml_data[list_to_compare]}
+
+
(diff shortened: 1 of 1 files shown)


scripts/yamlcompare.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # Compare CDEs ############################################################
+    result_file.write("\nCommon Data Elements::\n")
+
+    dict1 = get_dict_of_items(data1, "cdes", "code")
+    dict2 = get_dict_of_items(data2, "cdes", "code")
+    compare_dicts(dict1, dict2, result_file)
+
+    # Compare Survey Questions ##################################################
+    result_file.write("\nSurvey Questions::\n")
```
id2359, line 63: doing this breaks the csv'iness of the file
Might be better if we are going to use csv to  have

item category ( e.g. survey or section or cde) , item name ( e.g. the name of the survey, or section code ) , item field ( eg allow_multiple of the section ) ,  A value,  B value

as columns

It's a bit artificial to squeeze data into this shape but is consistent csv

jithendralal: Script modified to make simple csv with 5 columns
category, item, attribute, valueA, valueB

Sample output of the script: https://github.com/muccg/rdrf/issues/1323

Issue #1327, opened 2020-05-22
fix workflow for curators ( reset password ) when receiving login assistance email mtm


id2359:

jithendralal:
Related issue: https://github.com/muccg/rdrf/issues/1308

jithendralal:
To test on MTM:

**Patient:**
Register as a patient. 
Get locked out by 3 wrong password login attempts.
Use the link in the locked notification email.
Submit email.
Use the link in the verification email.
Submit Patient details.
Reset password and login.

**Curator:**
Get locked out by 3 wrong password login attempts.
Use the link in the locked notification email.
Submit email.
Use the link in the reset password email.
Reset password and login.

**Clinical:**
Get locked out by 3 wrong password login attempts.
Use the link in the locked notification email.
Submit email.
Receive instructions to contact admin.

PR #1333, opened 2020-05-27
rdrf #1327 workflow for curators added for resetting password

jithendralal:
The MTM Curator users were not allowed to reset password when they get locked out (due to wrong password attempts). Instead, they used to get an email with instruction to contact admin.

This was fixed in this by allowing Curator users to set a new password when they get locked out.
They get an email with a secure password reset link when they submit their email from the password_assistance form.

[password_workflow_before.pdf](https://github.com/muccg/rdrf/files/4686580/password_workflow_before.pdf)

[password_workflow_after.pdf](https://github.com/muccg/rdrf/files/4686578/password_workflow_after.pdf)

### DIFF
--- a/rdrf/rdrf/auth/__init__.py
+++ b/rdrf/rdrf/auth/__init__.py
@@ -8,4 +8,4 @@ def is_user_privileged(user):
 def can_user_self_unlock(user):
     return (
         getattr(settings, 'ACCOUNT_SELF_UNLOCK_ENABLED', False) and not
-        is_user_privileged(user) and not user.is_curator and not user.prevent_self_unlock)
+        is_user_privileged(user) and not user.prevent_self_unlock)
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/auth/__init__.py
```diff
@@ -8,4 +8,4 @@ def is_user_privileged(user):
 def can_user_self_unlock(user):
     return (
         getattr(settings, 'ACCOUNT_SELF_UNLOCK_ENABLED', False) and not
-        is_user_privileged(user) and not user.prevent_self_unlock)
+        is_user_privileged(user) and not user.is_curator and not user.prevent_self_unlock)
```
id2359, line 11: Won't the is_user_privileged function return True for curators? i.e. they're not patients etc

id2359: So adding an explicit check for is_curator seems redundant

Issue #1336, opened 2020-06-03
Integer cdes in survey seem to require min and max value


id2359:
The original CIC Breast cancer survey has a question for height where the cde originally had no max value - Trying to open the survey resulted in a traceback when creating the client_rep of the survey.

( For the spec in the case the cde was of datatype integer)
was trying to do int(None)

id2359:
So rather than enforce max and min on integer cdes ( so as not to break the surveys), the react frontend should handle the case where max or min are not set.

jithendralal:
It was decided to add a validation instead of setting default values.

PR #1360, opened 2020-06-08
rdrf #1336 Integer CDEs' spec min and max set to default values

jithendralal:
Integer CDEs' spec - min and max set to default values, 
if min_value and max_value are missing in CDE.

### DIFF
--- a/rdrf/rdrf/models/proms/models.py
+++ b/rdrf/rdrf/models/proms/models.py
@@ -159,10 +159,12 @@ class SurveyQuestion(models.Model):
                 "allow_multiple": self.cde.allow_multiple,  # allow for multiselect options
             }
         elif self.cde.datatype == 'integer':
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/models/proms/models.py
```diff
@@ -159,10 +159,12 @@ def _get_cde_specification(self):
                 "allow_multiple": self.cde.allow_multiple,  # allow for multiselect options
             }
         elif self.cde.datatype == 'integer':
+            min = int(self.cde.min_value) or 0
```
id2359, line 162: This will fail is min_value None

id2359: please test before pushing

jithendralal: thanks, corrected

Issue #1361, opened 2020-06-09
Add a min-max validation for survey question with Integer type CDE


jithendralal:
If an Integer type CDE is being used for the question, validate the min and max values.

PR #1362, opened 2020-06-09
rdrf #1361 Added validation for missing min and max values

jithendralal:
The validation is added to a survey question, if its CDE is of datatype integer.


rdrf/rdrf/models/proms/models.py
```diff
@@ -186,6 +189,11 @@ def clean(self):
                 # Check the cde is in one section only in the selected form.
                 self.validate_one_and_only_one_cde_exists()
 
+    def validate_min_max(self):
+        if not self.cde.min_value or not self.cde.max_value or self.cde.max_value < self.cde.min_value:
```
id2359, line 193: won't 0 as a min_value or 0 as a max_value now be impossible?

PR #1363, opened 2020-06-10
Custom Action Download/Polling View 

id2359:
The original custom action framework will time out on long running reports - this fixes that by adding a polling view which checks the async task created.

A download link becomes available when ready.

The result file is downloadable once ( it is deleted from the server at the time of download)


rdrf/rdrf/services/rest/views/api_views.py
```diff
@@ -379,11 +380,22 @@ def get(self, request, task_id):
...
                     if os.path.exists(filepath):
                         with open(filepath, 'rb') as fh:
-                            response = HttpResponse(fh.read(),
-                                                    content_type=content_type)
-                            response['Content-Disposition'] = "inline; filename=%s" % filename
-                            return response
-            return Http404
+                            file_data = fh.read()
+
+                        p = Popen("rm %s" % filepath, shell=True)
```
jithendralal, line 385: use of shell=True seems insecure

rdrf/rdrf/models/definition/models.py
```diff
@@ -2017,11 +2016,20 @@ def _generate_input_form_class(self, inputs):
...
+                from rdrf.forms.dynamic.fields import IsoDateField
+                klass = IsoDateField
+                widget = forms.DateInput(attrs={'class': 'datepicker'},
+                                         format='%dd-%mm-%YY')
+                kwargs["widget"] = widget
+                kwargs["input_formats"] = ["%d-%m-%Y"]
             else:
-                raise NotImplementedError("don't support yet")
-            return klass(label=label)
+                raise NotImplementedError("don't non date fields yet")
```
jithendralal, line 2030: you meant "not supported" but missed the word

id2359: ok

Issue #1358, opened 2020-06-05
ICHOMBC Survey: Follow up PROMS - health today widget


jithendralal:
**(Leslie to be consulted for this)**
The Health Today Widget is used for all survey questions with integer IDEs:
eg. your health past week, your quality of life past week, height, weight
The spacing of scale marks is not proper, the black box in the middle to the left of the scale shows "your health today"(hard coded) in all cases 
The widget to be made better configurable to suit all these cases.

jithendralal:
1. The marks on the scale are currently hard coded 10, 20, ....90.
2. The lowest and highest values are - the min_value and max_value of CDE
3. the black box in the middle can be made to show cde.name (currently hard coded "YOUR HEALTH TODAY") - but if the cde.name is very long, then it will not be displayed fully.
4. the labels - the best health you can imagine / the worst health you can imagine - are currently hard coded - to make this dynamic - some value in cde has to be used???

jithendralal:
```If the min_value=100 and max_value=0, currently the slider will not move ```

![image](https://user-images.githubusercontent.com/1198456/84334650-70f17200-abc5-11ea-93ae-54351074421e.png)

PR #1365, opened 2020-06-10
rdrf #1358 widget_config added to survey question

jithendralal:
Survey questions with integer CDEs were showing slider by default.
This is now controlled by a widget_config in the survey question.
(min_label, max_label, box_label are the parameters for slider widget)
The slider marks are calculated based on the min_value and max_value of integer CDE.

If the widget_config is not present,
then survey question with integer CDE will be rendered as an input box.


rdrf/rdrf/frontend/src/pages/proms_page/logic.ts
```diff
@@ -34,13 +34,22 @@ interface IntegerDatatype {
     min: number,
 }
 
+
+interface IntegerWidget {
+    widget_name: string,
```
id2359, line 39: try just widget_name here

rdrf/rdrf/frontend/src/pages/proms_page/logic.ts
```diff
@@ -34,13 +34,22 @@ interface IntegerDatatype {
     min: number,
 }
 
+
+interface IntegerWidget {
```
id2359, line 38: WidgetSpecInterface not IntegerWidget

Issue #1379, opened 2020-06-19
Integer survey question, if slider, to validate min max, else not to


jithendralal:
The the min and max of CDE are required if it is rendered as a slider in survey
If it is rendered as an input box, then min and max can be absent

PR #1382, opened 2020-06-24
rdrf #1379 Validation modified for Integer survey question

jithendralal:
If the integer survey question uses a widget config, then min_value and max_value of CDE made mandatory.

### DIFF
--- a/rdrf/rdrf/models/proms/models.py
+++ b/rdrf/rdrf/models/proms/models.py
@@ -205,9 +205,11 @@ class SurveyQuestion(models.Model):
                 self.validate_one_and_only_one_cde_exists()
 
     def validate_min_max(self):
-        if self.cde.min_value is None or self.cde.max_value is None:
-            raise ValidationError(
-                f"[{self.cde.code}] The min and max values are not properly set in CDE")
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/models/proms/models.py
```diff
@@ -205,9 +205,11 @@ def clean(self):
                 self.validate_one_and_only_one_cde_exists()
 
     def validate_min_max(self):
-        if self.cde.min_value is None or self.cde.max_value is None:
-            raise ValidationError(
-                f"[{self.cde.code}] The min and max values are not properly set in CDE")
+        # the slider widget used (in surveys) for integer CDEs require min and max values
+        if self.widget_config is not None and self.widget_config.strip() != "":
+            if self.cde.min_value is None or self.cde.max_value is None:
```
id2359, line 210: Add widget_name to config ( so the slider is called "slider" ) - check for this here 

e.g. widget_config = json.loads(widget_config_json)
widget_name = widget_config.get("widget_name", "")
if widget_name == "slider"

jithendralal: The modification was done.

Issue #1393, opened 2020-06-26
management command to update the registry version 


id2359:
( this is attached to the registry model object )

PR #1396, opened 2020-06-26
rdrf #1393 Management command added for setting the registry version.

jithendralal:
- The management command to display the current registry version and to accept an input of the new version from the user, to be set to the registry

### DIFF
--- a/rdrf/rdrf/management/commands/set_registry_version.py
+++ b/rdrf/rdrf/management/commands/set_registry_version.py
@@ -1,3 +1,4 @@
+import sys
 from django.core.management import BaseCommand
 from rdrf.models.definition.models import Registry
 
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/management/commands/set_registry_version.py
```diff
@@ -0,0 +1,15 @@
+from django.core.management import BaseCommand
+from rdrf.models.definition.models import Registry
+
+
+class Command(BaseCommand):
+    help = "Set the registry version"
+
+    def handle(self, *args, **options):
```
id2359, line 8: Make more specific as we do have the possibility of having more that one registry on a site.

rdrf/rdrf/management/commands/set_registry_version.py
```diff
@@ -6,10 +6,14 @@ class Command(BaseCommand):
...
-        print("\nATTENTION: You are about to change the version of Registry")
-        for registry in Registry.objects.all():
-            current_version = registry.version or "Not set"
-            print(f"\nCurrent version of {registry.name} is: {current_version}")
-            new_version = input("Enter new version: ")
-            registry.version = new_version
-            registry.save()
+        print("\nThe current registry versions:\n")
+
+        registries = Registry.objects.all().values("name", "code", "version").order_by("code")
```
id2359, line 11: Better to use same non-interactive pattern ( option parser) as for the other commands.

jithendralal: modified as suggested

rdrf/rdrf/management/commands/set_registry_version.py
jithendralal: modified as suggested

PR #1424, opened 2020-07-06
1417 integer min max

jithendralal:
Integer min and max validation added.
- for integer questions without widget config, if the integer cde has min and max set

### DIFF
--- a/rdrf/rdrf/models/proms/models.py
+++ b/rdrf/rdrf/models/proms/models.py
@@ -182,7 +182,7 @@ class SurveyQuestion(models.Model):
             if widget_spec:
                 ui = "integer-slider"
             params = {"type": "text"}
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/models/proms/models.py
```diff
@@ -181,13 +181,12 @@ def _get_cde_specification(self):
             ui = "integer-normal"
             if widget_spec:
                 ui = "integer-slider"
+            params = {"type": "text"}
+            if self.cde.min_value and self.cde.max_value:
```
id2359, line 185: distinguish  between is None and is zero

zero is allowed as min or max

jithendralal: changes made

PR #1453, opened 2020-07-20
1441 bcca report

jithendralal:
The BCCA Report related changes were made.
The BCCA report uses the type Patient Status Report
- the get handler was modified to handle the report request and to generate the report (since the BCCA report does not have inputs)

### DIFF
--- a/rdrf/rdrf/services/io/actions/patient_status_report.py
+++ b/rdrf/rdrf/services/io/actions/patient_status_report.py
@@ -92,6 +92,7 @@ class ReportGenerator:
         self.data = self._get_all_data()
         self.context_form_group = self._get_context_form_group()
         self.report = None
+        self.has_valid_filter = False
         self._setup_inputs()
 
     def _setup_inputs(self):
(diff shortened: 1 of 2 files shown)


rdrf/rdrf/services/io/actions/patient_status_report.py
```diff
@@ -393,12 +393,16 @@ def _completed(self, patient_model, form_name, data, percentage=False):
...
-            if self.has_no_filter or self._include_patient(patient_model):
-                yield patient_model
+        if not self.has_filter:
+            return Patient.objects.filter(rdrf_registry__code__in=[self.registry_model.code],
+                                          working_groups__in=user_working_groups).iterator()
+        else:
+            def patient_iterator():
+                for patient_model in Patient.objects.filter(rdrf_registry__code__in=[self.registry_model.code],
+                                                            working_groups__in=user_working_groups):
+                    if self._include_patient(patient_model):
```
id2359, line 403: if we can push out the test for self.start_date and end_date outside the loop would be good

rdrf/rdrf/services/io/actions/patient_status_report.py
```diff
@@ -393,12 +393,16 @@ def _completed(self, patient_model, form_name, data, percentage=False):
...
-
-        for patient_model in Patient.objects.filter(rdrf_registry__code__in=[self.registry_model.code],
-                                                    working_groups__in=user_working_groups):
-
-            if self.has_no_filter or self._include_patient(patient_model):
-                yield patient_model
+        if not self.has_filter:
+            return Patient.objects.filter(rdrf_registry__code__in=[self.registry_model.code],
+                                          working_groups__in=user_working_groups).iterator()
+        else:
```
id2359, line 399: just return query set

Issue #1618, opened 2020-11-06
Add alternative use case scenario for fh feature file ( edit patient)


id2359:
Add aloe test / step to test for correct error message on saving an address _without_ a country selected

Should see this in red message window

Please correct errors below:
Patient Address country: This field is required

Extra points if you can make the message wiindow check reusable (  green ( good)  or red (bad) should be part of the step or helper function config)

PR #1619, opened 2020-11-06
rdrf #1618 Added feature to test for missing country & step to check for alert boxes

APMCCG:
Added new step to `steps.py` to check for alert message boxes, and new feature `fh_add_patient_empty_address_country.feature` to test said step by checking for the error alert message when attempting to save a patient address without entering the country.


id2359 (changes requested):
Couldn't we have a step like

Then I should see an error message saying "blah"  ?
because

Then I should see an error message
And I should see "xyz" 

would be true if xyz is on trhe page but not in the error message part

( I realise it will in fact work in this case , because the error message is specific enough )

id2359 (approved):

Issue #1625, opened 2020-11-19
RDRF: Add aloe test for password reset form


APMCCG:
This ticket is to track work on adding sufficient steps to `steps.py` and relevant feature files to test if the password reset form completes successfully instead of causing a run-time error.

PR #1630, opened 2020-11-27
Rdrf #1625 Added password reset workflow test

APMCCG:
Added new steps to `steps.py` and created `ciccrc_passwordreset.feature` to test that the workflow for sending an email to reset a user's password progresses as expected and causes no run-time errors.


id2359 (changes requested):
See if you can avoid hardcoding any target urls

id2359 (approved):

id2359 (approved):

PR #1652, opened 2021-01-14
rdrf #1642 Added whitelist to script to ignore views that do not require security measures

APMCCG:
Added a list of views to `check_views.py` that neither have decorators on their get/post functions nor inherit from the `LoginRequiredMixin` class - these views do not require/will not function correctly with these security measures.


id2359 (changes requested):
1. rename that _str variable I usually use plurals for variables that are lists , i.e. superclasses 
2. Usually a good idea to open files with the with keyword:

with open(filename) as filehandle:
        data = filehandle.readlines()
# now filehandle closed

id2359 (approved):

Issue #1744, opened 2021-04-14
RDRF: React frontend - unexpected action payload


APMCCG:
The `submitAnswers` action, used in the reducer to send survey data to RDRF, is currently used as the "onClick" event handler for the "Submit Answers" button on the last page of the PROMS surveys in the frontend. This causes the emitted action to have an empty event object attached to it as a payload - this payload is not being used by the reducer, and has caused issues in React testing.

id2359:
Make payload null like the other actions that don't use a payload.

PR #1753, opened 2021-04-20
RDRF #1744 Removed submitAnswers payload

APMCCG:
- submitAnswers action was being used as event handler
- Added "handleSubmit" event handler to fire submitAnswers
- Adjusted testing to account for changes
- Performed small linting and yarn build fixes

### DIFF
--- a/rdrf/rdrf/frontend/src/app/index.tsx
+++ b/rdrf/rdrf/frontend/src/app/index.tsx
@@ -65,6 +65,10 @@ class App extends React.Component<AppInterface, object> {
         }
     }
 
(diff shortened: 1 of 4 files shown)


rdrf/rdrf/frontend/src/app/index.tsx
```diff
@@ -65,6 +65,10 @@ class App extends React.Component<AppInterface, object> {
         }
     }
 
+    public handleSubmit = (event) => {
+        this.props.submitAnswers();
```
id2359, line 69: good

id2359 (approved):

Issue #1760, opened 2021-04-28
RDRF: Session timeout cannot be disabled


APMCCG:
In RDRF 6.3.18. Follow up to #1759.
It is currently not possible to disable the timeout for a login session due to the way `django-session-security` is currently implemented. This issue can be addressed by making the use of the module conditional on the `SESSION_SECURITY_ENABLE` environment variable.

PR #1763, opened 2021-04-29
Rdrf #1760 - Make system respond to session security env var

APMCCG:


id2359 (changes requested):
Lets default to True - so we're secure by default

id2359 (approved):

Issue #1754, opened 2021-04-20
RDRF: Refactor worst (according to radon ) modules


APMCCG:
The Importer class in `rdrf/rdrf/services/io/defs/importer.py` requires refactoring to reduce Cyclomatic Complexity (as calculated by Radon), particularly in the `_create_registry_objects` method. This is also the case in `rdrf/rdrf/views/form_view.py` with the `post` method.

APMCCG:
`_create_registry_objects` has been refactored to reduce complexity, but further refactoring is required in other parts of the code.

PR #1764, opened 2021-04-29
RDRF #1754 refactor worst modules

APMCCG:


rdrf/rdrf/services/io/defs/importer.py
```diff
@@ -540,55 +540,30 @@ def _create_forms(self, registry, imported_forms):
...
-            logger.info("imported custom actions OK")
+        remaining_objects = [
+            "demographic_fields",
+            "complete_fields",
+            "reports",
+            "cde_policies",
+            "context_form_groups",
+            "email_notifications",
+            "consent_rules",
+            "surveys",
```
id2359, line 551: yeah this seems ok - or you could use some other data structure to store a bit more config for the exceptional cases, a dict maybe

id2359 (approved):

Issue #1228, opened 2020-02-21
FH/OPHG: Incomplete data prefill on patient demographics page


gmaloneccg:
In FH, when changing the structure of family linkage, in particular swapping index patients, the Sex, Location, and Living status fields for relatives are missing data on the index patient's demographics page.

Reprex:
1. Create new patient and successfully save (do not add relative at this stage)
2. Add relative to the new patient via demographics, and create relative patient
3. Navigate to the family linkage page:
3.1. Move relative to the index slot and save
3.2. Move old index back to index slot and save
4. Check relatives listed on index patient's demographics page -- Sex, Location and Living status of the relative should be empty

APMCCG:
Issue is still presenting in RDRF 6.3.19.
Steps to recreate are the same as original ticket.

PR #1781, opened 2021-05-11
RDRF #1228 - Added missing data to relative patient

APMCCG:
- _change_index method of FamilyLinkageManager did not use sex,
    living status, or address of original index when converting to
    relative
- Added data assignments between original index and new relative
- Used Pycountry to set location

### DIFF
--- a/rdrf/rdrf/views/family_linkage.py
+++ b/rdrf/rdrf/views/family_linkage.py
@@ -10,8 +10,7 @@ from django.db import transaction
 from django.core.management import call_command
 
 from rdrf.models.definition.models import Registry
-from registry.patients.models import Patient, PatientRelative, PatientAddress
-import pycountry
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/views/family_linkage.py
```diff
@@ -171,6 +172,10 @@ def _change_index(self, old_index_patient, new_index_patient):
             elif relative_dict["class"] == "Patient":
                 # index 'demoted' : create patient rel object
                 patient = Patient.objects.get(pk=relative_dict["pk"])
+                try:
+                    patient_location = PatientAddress.objects.get(patient=patient).country
```
id2359, line 176: Guard against MultipleObjectsReturned. Get the home address

id2359 (changes requested):

id2359 (approved):

Issue #1807, opened 2021-05-27
PROMS URLs included on non-PROMS site systems


APMCCG:
In RDRF 6.3.25.
URLs that are only meant for use on the PROMS site are also being registered for regular CIC and non-CIC sites in rdrf/rdrf/urls.py.

id2359:
3 lists of url patterns need to be setup ( not two)

PR #1810, opened 2021-05-31
rdrf #1807 Reworked URL patterns in rdrf/rdrf/urls.py

APMCCG:
- Removed unnecessary PROMS URLs from normal patterns
- Removed explorer/report URLs from normal patterns
- Added extra conditional URLs for CIC_DEV and NORMAL system roles

### DIFF
--- a/rdrf/rdrf/urls.py
+++ b/rdrf/rdrf/urls.py
@@ -104,41 +104,6 @@ proms_site = [
 
 ]
 
-proms_patterns = [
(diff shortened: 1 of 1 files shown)


id2359 (changes requested):
normal patterns ( ie normal sites like FH) do need the explorer report urls

CIC site or proms doesn;t make use of these though ( we use custom actions for reports)

PR #1870, opened 2021-07-13
1869 csp config

jithendralal:
csp is disabled by default
set the env var to enable csp

- ENABLE_CSP=1

### DIFF
--- a/rdrf/rdrf/settings.py
+++ b/rdrf/rdrf/settings.py
@@ -159,7 +159,7 @@ MESSAGE_TAGS = {
 # shows up messages addressed to other users.
 MESSAGE_STORAGE = 'django.contrib.messages.storage.session.SessionStorage'
 
-MIDDLEWARE = (
+MIDDLEWARE = [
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/settings.py
```diff
@@ -189,12 +188,13 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'django_user_agents.middleware.UserAgentMiddleware',
-    'csp.middleware.CSPMiddleware',
-)
+]
 
 if SYSTEM_ROLE == SystemRoles.CIC_PROMS:
     MIDDLEWARE = PROMS_MIDDLEWARE
```
id2359, line 194: default to csp enabled for proms , false for non-proms

id2359 (changes requested):

rdrf/rdrf/settings.py
jithendralal: Done this.

Issue #1868, opened 2021-07-12
wording next to radio options in react component of radio select


id2359:
Is it possible for some of the wording to be next to the survey numbers rather than have the explanations for scoring above the buttons?

e.g

Ο 0         No trouble at all

Ο 1         Mild

Ο 2

Ο 3

Ο 4         Moderate

Ο 5           

Ο 6

Ο 7         Severe

Ο 8

Ο 9

Ο 10       Worst I can imagine.

id2359:
![Uploading Screenshot_2021-07-12_14-31-21.png…]()

id2359:
![Screenshot_2021-07-12_14-31-21](https://user-images.githubusercontent.com/1163895/125241409-f80b9f00-e31d-11eb-9916-e9ea7ac1a25a.png)

APMCCG:
If the instructions need to be removed from the question title following this update, a definition change will need to be made.

APMCCG:
Properly updating the MOST questions with the scoring explanations next to the options is currently not possible without creating new React widgets & components, as the data being passed from the site system to the surveys does not contain all of the requisite information to do so.

id2359:
I think this would be better done as a separate React component.

id2359:
Not high priority

PR #1877, opened 2021-07-14
RDRF #1868 Added addOptionText method to Question component

APMCCG:
- Substituted addOptionText for label text in renderOptions
- Different sets of option text depending on order of options

### DIFF
--- a/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
+++ b/rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
@@ -184,6 +184,63 @@ class Question extends React.Component<QuestionInterface, object> {
         return handle;
     }
 
+    public addOptionText(question: any, optionText: string) : string {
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/frontend/src/pages/proms_page/components/question.tsx
```diff
@@ -184,14 +184,71 @@ class Question extends React.Component<QuestionInterface, object> {
...
+                        break;
+                }
+            } else {
+                switch(optionText) {
+                    case "10":
+                        returnText += "\t - Best possible";
+                        break;
+                    case "9":
+                        returnText += "\t - Very good";
+                        break;
```
id2359, line 219: Find alternative to switch

id2359 (changes requested):

id2359:
I think for this one , we're better off with a new widget. If the researchers want it we'll invest the time but not now.

Issue #1904, opened 2021-07-28
Optimise survey endpoint code where possible


id2359:
loading registry <-- if we can avoid a full load even better 
loading survey  <------ this seems unnecessary

This should reduce the time taken to post a request.
Best is to avoid unnecessary Django queries at all

cache if related to definition 

but caching won't really help if survey submissions intermittent

PR #1905, opened 2021-07-28
rdrf #1904 SurveyEndPoint - cached registry

jithendralal:


rdrf/rdrf/services/rest/views/proms_api.py
```diff
@@ -34,6 +35,17 @@ def multicde(cde_model):
     return datatype == "range" and cde_model.allow_multiple
 
 
+def get_registry(code):
+    query_cache = caches["queries"]
+    key = f"Registry_{code}"
+    if key in query_cache:
+        return query_cache.get(key)
+    else:
```
id2359, line 43: can't we abstract this into a helper function in utils

ie
def get_rdrf_model(model):
     key = get_key(model)
     if key in cache etc

rdrf/rdrf/services/rest/views/proms_api.py
```diff
@@ -45,13 +57,9 @@ def post(self, request):
...
 
         try:
-            registry_model = Registry.objects.get(code=registry_code)
-
-            survey_model = Survey.objects.get(registry=registry_model,
-                                              name=survey_name)
-
-            survey_assignment = SurveyAssignment.objects.get(registry=survey_model.registry,
-                                                             survey_name=survey_model.name,
+            registry_model = get_registry(registry_code)
```
id2359, line 60: according to this https://docs.djangoproject.com/en/3.2/topics/db/queries/#queries-over-related-objects
we only need primary key 
so maybe get id of registry from code :
id = Registry.objects.get(code=code).values('id')

then

survey_assignment = SurveyAssignment.objects.get(registry=id

id2359 (changes requested):

rdrf/rdrf/services/rest/views/proms_api.py
jithendralal: .values cannot be used with .get - can be applied to queryset

jithendralal: Registry.objects.filter(code="ICHOMCRC").values_list('id', flat=True)[0]

rdrf/rdrf/helpers/cache_utils.py
```diff
@@ -30,3 +31,29 @@ def wrapper(*args, **kwargs):
...
+            query_cache.set(key, result)
+            return result
+    return wrapper
+
+
+@use_query_cache_for_model
+def get_rdrf_model_id(class_name, value):
+    model_class = apps.get_model("rdrf", class_name)
+    try:
+        return model_class.objects.filter(id=value).values_list('id', flat=True)[0]
```
id2359, line 57: But we don;t need to try id first

rdrf/rdrf/helpers/cache_utils.py
```diff
@@ -30,3 +31,29 @@ def wrapper(*args, **kwargs):
...
+    except Exception:
+        return model_class.objects.filter(code=value).values_list('id', flat=True)[0]
```
id2359, line 59: this bit is ok - as we need id from code

id2359 (changes requested):
Just fix up unnecessary clause where try to get id using value as id

id2359 (approved):

Issue #1925, opened 2021-08-30
Management command to re-populate redis umrns


id2359:

jithendralal:
This management command requires the following:

'blackboard' in settings:
```
CACHES['blackboard'] = {
    "BACKEND": "django_redis.cache.RedisCache",
    "LOCATION": env.getlist("cache", ["redis://rediscache:6379/9"]),
    "TIMEOUT": 3600,
    "OPTIONS": {
        "CLIENT_CLASS": "django_redis.client.DefaultClient"
    },
    "KEY_PREFIX": "rdrf_if_"
}
```
docker_compose_common.yml
```
  rediscache:
    image: redis:3-alpine
    command: ["redis-server", "--appendonly", "yes"]
    environment:
      - REDIS_DISABLE_COMMANDS=FLUSHDB,FLUSHALL
    volumes:
      - ./data/redis:/data
    ports:
      - "6379:6379"
```

PR #1936, opened 2021-09-08
rdrf #1925 management command to populate umrns in redis

jithendralal:
- added management command to populate the set of umrns in redis

### DIFF
--- a/rdrf/rdrf/management/commands/populate_redis_umrns.py
+++ b/rdrf/rdrf/management/commands/populate_redis_umrns.py
@@ -1,0 +1,36 @@
+import sys
+from django.core.management.base import BaseCommand
+from django_redis import get_redis_connection
+from registry.patients.models import Patient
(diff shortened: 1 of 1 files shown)


rdrf/rdrf/management/commands/populate_redis_umrns.py
```diff
@@ -0,0 +1,36 @@
...
+            self._print("Error: registry code required")
+            sys.exit(1)
+        try:
+            Registry.objects.get(code=registry_code)
+        except Registry.DoesNotExist:
+            self._print("Error: registry does not exist")
+            sys.exit(1)
+
+        conn = get_redis_connection("blackboard")
+        umrns = set(Patient.objects.filter(umrn__isnull=False).filter(
```
id2359, line 32: no need to make a set here because sadd will not dupes

id2359 (approved):

Issue #1938, opened 2021-09-09
Update client rep json 


jithendralal:
On making changes to the survey and saving the survey, 
the Client rep json is not updating to the latest state.

When a few changes were made by saving each change,
it is found that the json was updated to the previous state.

UPDATE:

Better we do this on synchronise to proms once only

id2359:
This may already be working for syncing anyway - test in cic dev

PR #1947, opened 2021-09-24
rdrf #1938 client rep updated on changing survey question

jithendralal:


id2359 (changes requested):
Let's refactor and do the calcuation once only on synchronise to the proms

PR #1976, opened 2021-10-07
Rdrf 1966

id2359:
Added a visual indicator that a field is external + explanation ( only on demographics so far).
Added a context processor to only show for those sites where a hub is enabled.


jithendralal:
Marking all fields that appear in HL7Mapping as 'External', looks good.

PR #1987, opened 2021-10-14
1978 hl7 mappings were added to exporter and importer

jithendralal:


rdrf/rdrf/services/io/defs/importer.py
```diff
@@ -844,7 +844,7 @@ def _create_hl7_mappings(self, registry):
         if "hl7_mappings" in self.data:
             hl7_mappings = self.data["hl7_mappings"]
             for hl7_mapping in hl7_mappings:
-                HL7Mapping.objects.create(**hl7_mapping)
+                HL7Mapping.objects.get_or_create(**hl7_mapping)
```
id2359, line 847: check just on the event_code oterwise we'll create new all the time

PR #2061, opened 2021-11-10
External demographics view

jithendralal:
- New base template for external demographic fields readonly form
- new view for external demographics form
- template for external demographics form
- displayed readonly label / value pairs for all demographic external fields
- made sidebar work on new form
- change link to current demographics if external demographics form used


rdrf/rdrf/views/patient_view.py
```diff
@@ -1153,3 +1176,104 @@ class QueryPatientView(View):
...
+        context = {}
+        patient_fields = self._patient_fields(user, registry, patient._meta.fields)
+
+        def get_label(field):
+            return field.verbose_name
+
+        def get_value(patient, field):
+            value = getattr(patient, field.name)
+            if field.name == "sex":
+                for choice in Patient.SEX_CHOICES:
```
id2359, line 1258: good

id2359 (changes requested):
Just remove my code to get the patient form class as it is now not being used,

PR #2080, opened 2021-11-16
2026 error messages

jithendralal:
- added error messages to message model

### DIFF
--- a/rdrf/intframework/views.py
+++ b/rdrf/intframework/views.py
@@ -128,6 +128,13 @@ class IntegrationHubRequestView(View):
         logger.info(f"getting hub response for umrn {umrn}")
         hub = self._get_hub(registry_model, user_model)
         if hub is None:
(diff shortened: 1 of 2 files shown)


rdrf/intframework/views.py
```diff
@@ -128,6 +128,13 @@ def _get_hub_response(self, registry_model, user_model, umrn: str) -> Optional[d
         logger.info(f"getting hub response for umrn {umrn}")
         hub = self._get_hub(registry_model, user_model)
         if hub is None:
+            result_model = HL7Message()
+            result_model.username = user_model.username
+            result_model.registry_code = registry_model.code
+            result_model.umrn = umrn
```
id2359, line 134: do we allow the content to be empty - might be better to add result_model.content = " no message" if not

id2359 (changes requested):

PR #2114, opened 2021-12-06
Rdrf 1986 treating

jithendralal:
- fixed hard coding of treating clinician and added a model dropdown lookups as an alternative


id2359 (changes requested):
good but we need to support the  old notation too PatientCentres

jithendralal:
Changes made for the earlier version of widgets work: DataSourceSelect:PatientCentres

This was tested locally:
by adding a CDE with widget DataSourceSelect:PatientCentres
and another CDE with widget DataSourceSelect and widget_config {"tag":"diabetic"}

